Issue #13, opened 2020-01-31
Investigate: integration with Jupyter


fabioz:

datakurre:
Seems to work by mostly following the instructions at https://github.com/krassowski/jupyterlab-lsp

```json
{
  "LanguageServerManager": {
    "language_servers": {
      "robotframework-lsp": {
        "version": 1,
        "argv": ["robotframework_ls", "-v", "--log-file=robotframework_ls.log"],
        "languages": ["robotframework", "plain"]
      }
    }
  }
}
```

![robotframework-lsp](https://user-images.githubusercontent.com/160447/73885441-708c4780-4870-11ea-88fe-5d66db2e7fc9.png)

I tried out with robotframework 3.2b1 and robotkernel GitHub master.

/cc @bollwyvl  (Any ideas, why I needed to add "plain" into the list of robotframework_ls languages to make jupyterlab-lsp to properly initialize? `jupyter-lsp==0.7.0b0`)

bollwyvl:
Ooh exciting!

The mime/language/extension mapping is very tenuous still. I think a bunch
of things are hard coded (and inside closures) on that side. Will get back
to you.

In an upcoming release, we'll probably change a lot of things to be more
implementation-specific (e.g. /lsp/rfls rather than /lsp/robotframework),
and try to lean more on lab content types.

fabioz:
@datakurre thanks for investigating!

I'll leave it open for now for comments -- although it seems the investigation already went well ;)

PR #16, opened 2020-02-06
Add binder

bollwyvl:
This adds a binder demonstrating integration with JupyterLab (#13) via `jupyter(lab)-lsp` (which is also under heavy development).

[Binder](https://mybinder.org/v2/gh/bollwyvl/robotframework-lsp/add-binder?urlpath=lab)

Key points:
- the `binder` folder contains all the stuff to "reproducibly" fire up a demo
  - reproducible in this case _definitely_ doesn't mean production-ready, nix-hardend, bit-for-bit reproducible, but it should continue to work for the foreseeable future
  - would probably want a demo notebook (or even a whole workspace) in there, too
- an `entry_point` is added to `setup.py`
  - it doesn't _force_ anybody to do anything, but will be available if someone _were_ to install jupyter_lsp (and the raft of stuff that goes with it) and not having to monkey around with paths
  - this spec will eventually be expanded to include all sorts of things (like configuration JSON schema) but presently is mostly about `argv`

@datakurre indeed you were right: plain-old `.robot` files work fine, but `.ipynb` doesn't without `plain`. I'll have a look at it :blush:: there's a _lot_ of magic going on there to even work as well as it does, and we have been pretty much concentrating on r and python notebooks vs their respective (or each others!) language servers.

Screenshot from 2020-02-05 21-07-04


fabioz:
@bollwyvl seems nice... do you think it'd be worth to add something to the README regarding this?

Also, I'll probably squash it when integrating (is that ok or would you like to keep some of that internal history?)

sahokas:
Nick, Asko, welcome - nice to see you here!

Regarding @fabioz question about squash vs keeping internal history: please feel free to proceed as you see fit, but just FYI - in Robocorp internal development we default to squash unless there is a specific reason to keep the history.

bollwyvl:
> add something to the README

Sure! I didn't add it because the URL wouldn't work yet, but the markdown would be:

```markdown
[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/robocorp/robotframework-lsp/master?urlpath=lab)
```
If you add a doc you'd like to open (preferably an `.ipynb`), the urlpath would just become, `?urlpath=lab/tree/docs/index.ipynb`. Having it open a whole workspace (multiple documents/parts of documents) is [another matter](https://github.com/ian-r-rose/binder-workspace-demo/).

> squash it when integrating

Nope, debugging-at-a-distance on Other People's Computers always leaves a nasty trail. Squash away!

PR #38, opened 2020-04-07
Updated tmLanguage to be more compliant with typical element types

Snooz82:
Hi @fabioz 

i updated the tmLanguage again to better fit to typical types.
The types have been a little wierd before.


fabioz:
Thanks for the update ;)

PR #40, opened 2020-04-08
Small contribution to VSCode Plugin

Snooz82:
- small improvements in language-configuration.json
 - small visual improvements to package.json (icon, background, displayname)
 - added keybinding for tab key to send 4 spaces (editable with robot.editor.4spacesTab)


Snooz82:
Hey @fabioz ,

i made some small things that should improve user experience with the VSCode plugin.

language configuration:
- added *bold* and _italic_ to the closingPairs and surroundingPairs

Small improved icon

package.json
- Added a Display Name to package JSON
   ``Robot Framework Language Server``
- Keybinding that uses tab key to type 4 spaces (best practice in robot framework)
- This keybinding can be controlled by the option robot.editor.4spacesTab (default is true)

i hope this is fine for you.

package.json
```diff
@@ -17,11 +17,20 @@
...
+    "galleryBanner": {
+        "theme": "dark",
+        "color": "#000000"
+    },
     "contributes": {
         "configuration": {
             "title": "Robot Framework Language Server Configuration",
             "type": "object",
             "properties": {
+                "robot.editor.4spacesTab": {
```
fabioz, line 29: Regarding the tabs vs spaces, vscode itself should handle that (we shouldn't need to cover this in the extension itself as the vscode default seems to be what we want anyways).

From https://code.visualstudio.com/docs/editor/codebasics:

- Indentation

  VS Code lets you control text indentation and whether you'd like to use spaces or tab stops. By default, VS Code inserts spaces and uses 4 spaces per Tab key. If you'd like to use another default, you can modify the `editor.insertSpaces` and `editor.tabSize` settings.

-- if the user has multiple languages and wants different settings, it should be handled as explained in: https://github.com/Microsoft/vscode/issues/23933

language-configuration.json
```diff
@@ -15,14 +15,18 @@
         ["[", "]"],
         ["(", ")"],
         ["\"", "\""],
-        ["'", "'"]
+        ["'", "'"],
+        ["*", "*"],
```
fabioz, line 19: I don't really understand why auto close `*` and `_` (testing here it seemed strange -- can you elaborate a bit on the scenario for this?)

Snooz82: This was just an idea because of the Documentation syntax. *bold* and _italic_ .

Maybe we can delete this from autoclose. You are right. This would be wierd everytime you press underscore somewhere and get two of them.

But surrounding pairs could be handy.

Snooz82: This was just an idea because of the Documentation syntax. *bold* and _italic_ .

Maybe we can delete this from autoclose. You are right. This would be wierd everytime you press underscore somewhere and get two of them.

But surrounding pairs could be handy.

package.json
Snooz82: This is only half of the truth.
VSCode always (and that can not be disabled) align the spaces to a multiple of the tab size.

That means if i have a keyword that has 8 characters and before are 4 spaces and you press tab, you get 4 spaces.
But if you have a keyword that has 3 characters and 4spaces ahead, you just get 1 space. 


`....Log.`
And
`....Log To Console..`
`....Run Keyword.`

And so on.
This is really annoying!!!
This 4spacesTag adds always exactly 4 spaces. Independent how long the line is, but only if the file is a Robot Framework file and if no text is selected.

I think this is a super cool feature.
We definitely can move this command to the typescript code later in a command listener but for now this is very useful.

And it can be deactivated via a checkbox.

language-configuration.json
Snooz82: Fixed

fabioz:
Thanks, for the pull request! Just merged ;)

PR #41, opened 2020-04-09
Bug fix for keyword highlighting

Snooz82:
Hi Fabio,
This belongs to the commit you already merged. 
in some Themes the toke variable.fuction.robot had no color assigned, so that the higher level color has been used.
I fixed it with setting keywords to "support.function.robot" which is normally used for language based functions like print(), but it does work so far.


fabioz:
Thanks for the fixes ;)

PR #43, opened 2020-04-10
Repository Key is object

Snooz82:
Hi Fabio,

just a very small fix.
i figured out that the "repository" element in Package.json shall be an object with type and url instead of a string.

Cheers
And Happy Eastern


fabioz:
Thanks for the fix. Just integrated.

Issue #52, opened 2020-04-25
Tab Completion not working on VSCode


JoneSabino:
Tab completion is not working on VSCode, even with the setting `"editor.tabCompletion": "on"` 

The suggestion shows up normally, but I need to press RETURN to auto complete not TAB, as usual.

Snooz82:
I will see if it can be fixed. I assume it has to do with the option Tab4Spaces.

fabioz:
@Snooz82 thanks for taking a look at it.

Snooz82:
This should be fixed by PR

fabioz:
Note: this is fixed for 0.1.1.

PR #62, opened 2020-05-05
FIXES: #52 Tab Completion not working

Snooz82:
It is now possible to use TAB to complete code snippets and also step between the arguments.


fabioz:
Thank you for the fix ;)

Issue #69, opened 2020-05-14
[Feature Request] Python Like Intellisense for Robot Framework For Loops, ELSE IF


Deep-Adhikary:
It Would be  helpful if place holder like intellisense is given for Robot Framework For Loops & ELSE IF  l
![image](https://user-images.githubusercontent.com/13049680/81898135-b70ee200-95d5-11ea-8458-7c6877af2093.png)
![image](https://user-images.githubusercontent.com/13049680/81898207-d9086480-95d5-11ea-90a1-b292afb76fe2.png)

Snooz82:
This is a nice idea!

At the moment i would think that these `code snippets` would be easy and super helpful.

i think these could be the code snippets:

FOR/IN RANGE
 ```
   FOR    ${i}    IN RANGE    ${START}    ${END}    ${STEPS}
        Log    ${i}
    END
```

FOR/IN
```
    FOR    ${element}    IN    @{LIST}
        Log    ${element}
    END
```

FOR/IN ENUMERATE
```
    FOR    ${index}    ${element}    IN ENUMERATE    @{LIST}
        Log    ${index}: ${element}
    END
```

FOR/IN ZIP
 ```
   FOR    ${number}    ${name}    IN ZIP    ${NUMBER_LIST}    ${NAME_LIST}
        Log    ${number}: ${name}
    END  
```

When IF/ELSE IF/ELSE will be introduced to robot in RF 4 this could then be added as well.

Snooz82:
@fabioz i just found out that these snippets are part of the VSCode Plugin.

i will prepare a PR for this later this day. So if you want to release, you may want to wait and include this PR. ;-)

Snooz82:
Preview:

![image](https://user-images.githubusercontent.com/41592183/82220895-8bfef800-991f-11ea-9d86-edff168aa0d3.png)
![image](https://user-images.githubusercontent.com/41592183/82220919-93be9c80-991f-11ea-8c1f-c1d2082336c0.png)

fabioz:
@Snooz82 yes, that'd work for VSCode, but it'd be nice to have them for other clients too (which doesn't really work if it's done that way).

I'm doing a reasonably big refactoring right now (with the idea of making it possible to do another extension using the same codebase), but as soon as I finish this I'll implement that (it should be straightforward to do).

Snooz82:
[...]

PR #71, opened 2020-05-18
Added Code Snippets to VSCode Plugin of FOR-Loops and Run Keyword If

Snooz82:
Solves #69 

This could be deleted when Language server itself supports this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -170,6 +170,12 @@
                     }
                 ]
             }
+        ],
+        "snippets": [
+            {
+                "language": "robotframework",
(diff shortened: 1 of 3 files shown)


fabioz:
I didn't merge initially due to some refactorings (so, it didn't apply cleanly), so, I decided to go on and just add it to the language server completions directly (so this is being closed in favor of https://github.com/robocorp/robotframework-lsp/commit/2a171a0b1ed8eba10e35dcd812c32d02016c85e2).

Issue #158, opened 2020-09-29
Add insertText to CompletionItem 


avaissi:
[jupyterlab-lsp extension](https://github.com/krassowski/jupyterlab-lsp) is using **insertText** attribute of the returned CompletionItem and ignoring the **text_edit** attribute. This is causing the extension to behave unexpectedly with RobotFramework LS

Needed:

Single field addition to returned CompletionItem.

For example:
robotframework-ls/src/robotframework_ls/impl/keyword_completions.py

**insertText=text_edit.newText**
```
return CompletionItem(
            keyword_found.keyword_name,
            kind=keyword_found.completion_item_kind,
            text_edit=text_edit,
            insertText=text_edit.newText,
            documentation=keyword_found.docs,
            insertTextFormat=InsertTextFormat.Snippet,
            documentationFormat=(
                MarkupKind.Markdown
                if keyword_found.docs_format == "markdown"
                else MarkupKind.PlainText
            ),
        ).to_dict()
```

fabioz:
Ok, it should be straightforward to add that, still, note that the `insertText` leaves too much up to interpretation from the client (which may end up not doing what is actually expected on some cases), so, ideally the jupyterlab-lsp extension would be fixed to do the right thing here...

i.e.: Spec (https://microsoft.github.io/language-server-protocol/specification) details on the related fields for CompletionItem
```
/**
 * A string that should be inserted into a document when selecting
 * this completion. When `falsy` the label is used.
 *
 * The `insertText` is subject to interpretation by the client side.
 * Some tools might not take the string literally. For example
 * VS Code when code complete is requested in this example `con<cursor position>`
 * and a completion item with an `insertText` of `console` is provided it
 * will only insert `sole`. Therefore it is recommended to use `textEdit` instead
 * since it avoids additional client side interpretation.
 */
insertText?: string;

/**
 * An edit which is applied to a document when selecting this completion. When an edit is provided the value of
 * `insertText` is ignored.
 *
 * *Note:* The range of the edit must be a single line range and it must contain the position at which completion
 * has been requested.
 */
textEdit?: TextEdit;
```

avaissi:
Added a suggestion for PR #160 

That's true that ideally the changes would be made into the [jupyterlab-lsp extension](https://github.com/krassowski/jupyterlab-lsp), but I'm not sure if I'm able to push needed changes there yet. However, it would help if the insertText would be populated by the LS server.

[...]

PR #160, opened 2020-09-29
insertText attribute value assigned for CompletionItem

avaissi:
Pull request suggestion for issue #158


fabioz:
Tests are failing because the expected data changed.

Most should be related to https://github.com/ESSS/pytest-regressions.

To force the update of the expected files it should be possible to call something as: `pytest -n auto --force-regen` (it should fail the time it regenerates the expected files and should pass on subsequent runs).

If you want I can update it here...

fabioz:
As I had to make changes for the tests, I ended up creating a new commit for the merge (so, was added in: https://github.com/robocorp/robotframework-lsp/commit/d9199729cf8d6ddda93a334dbdc4902f613d66ca).

PR #162, opened 2020-09-30
Added example configs

anton264:
Some example configs for run configurations and tasks.


fabioz:
Thanks for the docs. Just merged.

PR #331, opened 2021-04-22
refresh binder, update jupyter_lsp spec with install instructions

bollwyvl:
Hello there!

This hoists the binder config back up to root, and updates the spec to be detected by jupyter_lsp. 

It's testable here: [Binder](https://mybinder.org/v2/gh/bollwyvl/robotframework-lsp/restore-binder?urlpath=lab)

I've also updated the spec to not gobble up `plain`, as it's causing issues over on https://github.com/krassowski/jupyterlab-lsp/pull/493 because we kinda had the assumption under test that nobody would take it (whoops) and added the `install` instructions, which would propagate to our docs. There's a spot for extensions, so it could be nice to add `robocop` there.

It may also be worth looking at some other stuff in `setup.py`, like aligning the trove classifiers (which show up on PyPI) and the `python_requires` (which newer pips obey) to have a consistent story.  

Anyhow, the binder feels pretty good: selenium stuff works (with `headlessfirefox`, didn't try _too_ hard to get `chromium` working), debugger works, robocop works. Not blocking here, but hopefully we'll have a `robotkernel` soon which supports RF4 (which still has some nice interactive features over xeus-python)... and hopefully a working tests on `jupyterlab-lsp`! If the [notebook issue](https://github.com/krassowski/jupyterlab-lsp/issues/573) can be resolved, then we could probably take the spec management burden off of this repo altogether!


fabioz:
Just merged. Thanks for the fixes!

PR #400, opened 2021-07-18
Implemented completion for dictionary variable names

mardukbp:
Dictionary variable names can now be completed using the syntax:

```
${dict_var}
```

Currently they are only completed using the new syntax introduced in RF 4. which has a different meaning:

```
&{dict_var}

### DIFF
--- a/robotframework-ls/tests/robotframework_ls_tests/completions/test_variables_completions.py
+++ b/robotframework-ls/tests/robotframework_ls_tests/completions/test_variables_completions.py
@@ -383,8 +383,8 @@ def test_dictionary_variables_completions_with_dollar(
     from robotframework_ls.impl.completion_context import CompletionContext
     from robotframework_ls.impl import variable_completions
 
-    workspace.set_root("case10", libspec_manager=libspec_manager)
-    doc = workspace.get_doc("case10.robot")
+    workspace.set_root("case1", libspec_manager=libspec_manager)
+    doc = workspace.get_doc("case1.robot")
     doc.source = """*** Variables ***
 &{ROBOT}   Name=Robot Framework   Version=4.0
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by fabioz** [comment]
Thank you for the pull request... 

While the change itself seems fine, this pull request is missing some test-cases. Can you provide those?

It should be straightforward to create... You can add a new test in:
https://github.com/robocorp/robotframework-lsp/blob/master/robotframework-ls/tests/robotframework_ls_tests/completions/test_variables_completions.py (using one of the existing tests as a basis).

File: robotframework-ls/tests/robotframework_ls_tests/completions/test_variables_completions.py [hunk 1]
```diff
@@ -375,3 +375,45 @@ def test_variables_completions_in_variable_files_yaml(
...
+&{ROBOT}   Name=Robot Framework   Version=4.0
+
+***Test Cases***
+Test dictionary variable completion
+   Log to Console   ${ROB"""
+
+    completions = variable_completions.complete(
+        CompletionContext(doc, workspace=workspace.ws)
+    )
+    data_regression.check(completions)
```
fabioz, line 398: Note: the files generated by the data_regression must be committed.

mardukbp: How do I generate them?

fabioz: When you run the test it should automatically generate the data regression files (it's a `.yaml` which will be placed in a `test_variables_completions` folder next to the `test_variables_completions.py`).

Then, after re-running the test, if the results are different, it will fail (to regenerate you can just delete the related file and rerun -- if it's not there the file will be generated and the test will fail).

fabioz: Note that if you have full paths/uris related to your system, the test will fail due to such differences (so, it's possible that you have to change `completions` a bit to fix such references before sending it to the `data_regression.check`).

fabioz:
Just merged. Thank you for the pull request.

PR #407, opened 2021-07-30
find_definition now works for dictionary variables

mardukbp:
The current token might be `${dict}[key]`, which is not a variable. The token has to be filtered using a regex in order for `find_definition` to work for dictionary variables.


fabioz:
Just merged. Thanks for the fix!

PR #408, opened 2021-07-30
Complete dictionary items when the variable name includes spaces

mardukbp:
* Fixed the regex to match dictionary variables with spaces in their name.
* Added a corresponding test.


fabioz:
Just merged. Thanks for the fix!

PR #413, opened 2021-08-07
Report undefined variables as errors

mardukbp:
Variables that are not:

* defined in the Variables section of the current file or in a Resource file or in a Variables file
* defined as keyword arguments
* defined by assigning them to the return value of a keyword

are now reported as errors. Only variables passed as keyword arguments or used as dictionary values are analyzed.

### DIFF
--- a/robotframework-ls/src/robotframework_ls/impl/ast_utils.py
+++ b/robotframework-ls/src/robotframework_ls/impl/ast_utils.py
@@ -1,7 +1,8 @@
 import sys
 from typing import Iterator, Optional, List, Tuple, Any, Union
-
+import re
 import ast as ast_module
+from robot.api import Token
 from robocorp_ls_core.lsp import Error
 from robocorp_ls_core.robotframework_log import get_logger
 from robotframework_ls.impl.protocols import (
@@ -9,9 +10,11 @@ from robotframework_ls.impl.protocols import (
     NodeInfo,
     KeywordUsageInfo,
     ILibraryImportNode,
+    VariableUsageInfo,
 )
 
-
+_RF_VARIABLE = re.compile(r"([$|&|@]{[\w\s]+})")
+_RF_NUM_CONST = re.compile(r"([$]{\d+})")
 log = get_logger(__name__)
 
 
@@ -423,7 +426,7 @@ def get_documentation(ast) -> str:
 def iter_variable_assigns(ast) -> Iterator:
     from robot.api import Token
 
-    for stack, node in _iter_nodes(ast, recursive=False):
+    for stack, node in _iter_nodes(ast, recursive=True):
         if node.__class__.__name__ == "KeywordCall":
             for token in node.get_tokens(Token.ASSIGN):
                 value = token.value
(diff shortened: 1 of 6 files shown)


robotframework-ls/src/robotframework_ls/impl/ast_utils.py
```diff
@@ -1,17 +1,20 @@
 import sys
 from typing import Iterator, Optional, List, Tuple, Any, Union
-
+import re
 import ast as ast_module
+from robot.api import Token
```
fabioz, line 5: imports from `robot` should always be local imports, not top-level. 

-- note: in general, anything that isn't super-common should be a local import (in this case, the other imports are common, so, that's ok), and in particular, importing `robot` is super-slow, so, it should be done only when really needed.

robotframework-ls/src/robotframework_ls/impl/ast_utils.py
```diff
@@ -479,6 +488,25 @@ def create_keyword_usage_info(stack, node) -> Optional[KeywordUsageInfo]:
...
+
+    for stack, node in _iter_nodes(ast, recursive=True):
+        if node.__class__.__name__ in ["KeywordCall", "Variable"]:
+            for token in node.tokens:
+                usage_info = None
+                if token.type == Token.ARGUMENT:
+                    variable_name = _RF_VARIABLE.findall(token.value)
+                    if variable_name and not _RF_NUM_CONST.findall(variable_name[0]):
+                        usage_info = VariableUsageInfo(tuple(stack), node, token, variable_name[0])
+                        if usage_info is not None:
```
fabioz, line 506: I think that this `if usage_info is not None` is not needed here (since the instance was just created).

fabioz (changes requested):
I think undefined variables are highly susceptible to being false positives (because robot framework has many ways of specifying variables and its namespace isn't always local).

So, I think we should have some way of disabling this.

We can initially do this just through some environment variable (say `ROBOTFRAMEWORK_LS_REPORT_UNDEFINED_VARS`), where the truthy values are `True,1,true` and falsy values are `False,0,false` and later (in a different pull request) expand this to be handled through a section in a `pyproject.toml` (close to what `Robocop`: https://robocop.readthedocs.io/en/latest/user_guide.html is doing).

As a note, other ci tests are failing with this change...

robotframework-ls/src/robotframework_ls/impl/code_analysis.py
```diff
@@ -130,4 +161,28 @@ def collect_analysis_errors(completion_context):
             if len(errors) >= MAX_ERRORS:
                 # i.e.: Collect at most 100 errors
                 break
+
+    argument_names = list(ast_utils.iter_keyword_arguments_as_str(ast))
+    variable_assigns = list(ast_utils.iter_variable_assigns_as_str(ast))
+    variables_collector = _VariablesCollector()
+    collect_variables(completion_context, variables_collector)
+
+    for variable_usage_info in ast_utils.iter_variable_usage_tokens(ast):
```
fabioz, line 170: Another thing that occurred to me is that the scoping may not be correct here: in this implementation, when an argument is available for some keyword it seems it's made automatically available for all keywords for a match, which isn't correct. 

The implementation should probably be going into a Keyword and then collecting the variables which are valid just for that scope. Likewise, if a variable is used in a resource import, it needs to be defined globally and not inside some keyword import...

fabioz:
Closing as this was implemented in a different change.

As a note, this feature is now available in a pre-release (0.42.2).

See: https://github.com/robocorp/robotframework-lsp/blob/master/robotframework-ls/docs/changelog.md for all that's changed in the pre-release.

For this particular feature given that RF makes it easy to create variables in multiple ways, I'm curious on testing to filter out possible false positives.

PR #426, opened 2021-08-24
Avaissi/add inspector

avaissi:


robocorp-code/bin/create_env/conda.yaml
```diff
@@ -5,8 +5,6 @@ dependencies:
   - python=3.8
   - pip=20.1
   - pip:
+    - robocorp-inspector==0.3.2
     - robotframework
```
avaissi, line 9: @fabioz the robotframework is probably not a required dependency for the environment?
I could remove that and let the robocorp-inspector handle its dependencies

fabioz (changes requested):
I've tried to use it to edit an existing locator (i.e.: clicking at the edit button for the existing locator in the project below) and it gave an error.

```
Traceback (most recent call last):
  File "c:\temp\robohome\holotree\b1f3c244e_dc412287\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\temp\robohome\holotree\b1f3c244e_dc412287\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "c:\temp\robohome\holotree\b1f3c244e_dc412287\lib\site-packages\inspector\cli.py", line 117, in <module>
    run()
  File "c:\temp\robohome\holotree\b1f3c244e_dc412287\lib\site-packages\inspector\cli.py", line 107, in run
    app.edit(args.name)
  File "c:\temp\robohome\holotree\b1f3c244e_dc412287\lib\site-packages\inspector\app.py", line 35, in edit
    locator = self.ctx.load_locator(name)
  File "c:\temp\robohome\holotree\b1f3c244e_dc412287\lib\site-packages\inspector\context.py", line 83, in load_locator
    self.database.load()
  File "c:\temp\robohome\holotree\b1f3c244e_dc412287\lib\site-packages\inspector\database.py", line 82, in load
    self._create_cache()
  File "c:\temp\robohome\holotree\b1f3c244e_dc412287\lib\site-packages\inspector\database.py", line 61, in _create_cache
    locator["screenshot"] = load_image(locator["screenshot"])
  File "c:\temp\robohome\holotree\b1f3c244e_dc412287\lib\site-packages\inspector\database.py", line 53, in load_image
    with open(root / path, "rb") as infile:
OSError: [Errno 22] Invalid argument: 'x:\\vscode-robot\\local_test\\minimal\\iVBORw0KGgoAAAANSUhEUgAAARAAAABcCAYAAACm5+q2AAAAAXNSR0IArs4c6QAAIABJREFUeJztnXd4FVXawH9nZm5LSKOTEIogTQRBihUU14IFC1hWd62rq7K23U9dy9rWspa1lxXsuyIgShFxFUGwoEjvvScQCCSk3jYz5

...

```
[minimal.zip](https://github.com/robocorp/robotframework-lsp/files/7041249/minimal.zip)

-- Actually, not only the edit is failing, even adding a new locator is failing... 

Also, if the `locators.json` doesn't exist, trying to add a new locator is giving an error saying that it expected the locators.json to exist (I think it could just create it if it's not there...)

robocorp-code/bin/create_env/conda.yaml
fabioz: Actually I think we should keep the `robotframework` dependency because that's also the env that's used by default when the user uses the `Robot Framework Language Server` without using robots.

We should probably pin it to the latest `4.x` version thought...

fabioz (changes requested):
After creating a new `locators.json`, it seems to work, but then, when hovering over the screenshot field, the contents of the screenshot are no longer properly shown (this is done in `robocorp_code.robocorp_language_server.RobocorpLanguageServer.m_text_document__hover` -- it actually does seem it has support for relative paths, but then, it's not working for me with this version, but I haven't investigated why.

As a note, it seems that the locators.json no longer has support for the screenshot as a base64 image (that should probably be backward compatible to be able to load it, so that we don't break existing locators -- which is probably the case for existing users of vscode).

robocorp-code/bin/create_env/robot.yaml
```diff
@@ -7,3 +7,8 @@ tasks:
 artifactsDir: output
 
 condaConfigFile: conda.yaml
+environmentConfigs:
+  - environment_windows_amd64_freeze.yaml
```
fabioz, line 11: I wasn't aware of this, so, this is mostly a question from me trying to discover how this works... does this mean that on windows/amd64 it will use `environment_windows_amd64_freeze.yaml` instead of `conda.yaml`?

robocorp-code/src/robocorp_code/robocorp_language_server.py
```diff
@@ -1064,75 +1064,6 @@ def _resolve_interpreter(self, params=None) -> ActionResultDict:
         # i.e.: no error but we couldn't find an interpreter.
         return {"success": True, "message": "", "result": None}
 
-    @command_dispatcher(
```
fabioz, line 1067: I think that we no longer use the `robocorp_code.locators_db` module either, so, that module should probably be removed (as well as its test).

And the `_LocatorsInThreadAPI` class in this module seems to be unused too...

robocorp-code/vscode-client/src/inspector.ts
```diff
@@ -0,0 +1,198 @@
...
+async function startInspectorCLI(pythonExecutable: string, args: string[], environ?: { [key: string]: string }): Promise<ExecFileReturn> {
+    const inspectorCmd = ['-m', 'inspector.cli'];
+    const completeArgs = inspectorCmd.concat(args)
+    return execFilePromise(
+      pythonExecutable, completeArgs,
+      {env: {...process.env, ...environ}}
+    );
+}
+
+const getInspectorPythonInfoUncached = async (): Promise<InterpreterInfo | undefined> => {
```
fabioz, line 117: This shouldn't be needed (I guess this is some leftover... you're already using `ROBOCORP_GET_LANGUAGE_SERVER_PYTHON_INFO` in this module, so the `ROBOCORP_GET_INSPECTOR_PYTHON_INFO` can probably be completely removed (I can't see any usage of it).

robocorp-code/vscode-client/src/inspector.ts
```diff
@@ -0,0 +1,198 @@
...
+    let rccLocation = await getRccLocation();
+    if (!rccLocation) {
+        return;
+    }
+    let robotYaml = getExtensionRelativeFile('../../bin/create_inspector_env/robot.yaml');
+    if (!robotYaml) {
+        return;
+    }
+
+    async function createInspectorEnv(progress: Progress<{ message?: string; increment?: number }>): Promise<ExecFileReturn> | undefined {
```
fabioz, line 127: This should also be removed.

robocorp-code/vscode-client/src/inspector.ts
```diff
@@ -0,0 +1,198 @@
...
+      inspectorLaunchInfo.environ);
+    OUTPUT_CHANNEL.appendLine('Inspector CLI stdout:');
+    OUTPUT_CHANNEL.appendLine(launchResult.stdout);
+    OUTPUT_CHANNEL.appendLine('Inspector CLI stderr:');
+    OUTPUT_CHANNEL.appendLine(launchResult.stderr);
+}
+
+
+
+export async function getInspectorPythonInfo(): Promise<InterpreterInfo | undefined> {
```
fabioz, line 92: Remove this.

robocorp-code/vscode-client/src/inspector.ts
```diff
@@ -0,0 +1,198 @@
...
+import {getRccLocation} from './rcc'
+import {getExtensionRelativeFile, verifyFileExists} from './files'
+import {listAndAskRobotSelection} from './activities'
+import {getSelectedLocator, getSelectedRobot, LocatorEntry} from './viewsCommon';
+import {sleep, Timing} from './time'
+import {execFilePromise, ExecFileReturn} from './subprocess'
+import * as roboConfig from './robocorpSettings';
+import {OUTPUT_CHANNEL} from './channel';
+
+interface InterpreterInfo {
```
fabioz, line 13: Both declarations of `InterpreterInfo` in this module should probably be removed (this interface is already available in `extension.ts`, so if needed it can be imported from there).

robocorp-code/vscode-client/src/inspector.ts
```diff
@@ -0,0 +1,198 @@
...
+    additionalPythonpathEntries: string[];
+}
+
+interface InterpreterInfo {
+    pythonExe: string;
+    environ?: { [key: string]: string };
+    additionalPythonpathEntries: string[];
+}
+
+let _cachedInspectorPythonInfo: InterpreterInfo;
```
fabioz, line 25: _cachedInspectorPythonInfo shouldn't be needed...

robocorp-code/vscode-client/src/inspector.ts
```diff
@@ -0,0 +1,198 @@
...
+        robot = await listAndAskRobotSelection(
+          'Please select the Robot where the locators should be saved.',
+          'Unable to create locator (no Robot detected in the Workspace).'
+        );
+        if (!robot) return;
+    }
+    const robotYaml = robot.filePath;
+    locatorJson = path.join(robot.directory, "locators.json")
+    locatorJson = verifyFileExists(locatorJson) ? locatorJson : undefined;
+    let inspectorLaunchInfo = await commands.executeCommand(ROBOCORP_GET_LANGUAGE_SERVER_PYTHON_INFO) as InterpreterInfo | undefined;
```
fabioz, line 42: The actions are good when it's called from the language server or from another extension, but in this case I think we could call `getLanguageServerPythonInfo` directly (that way we don't need to cast anything).

robocorp-code/bin/create_env/conda.yaml
avaissi: I'll pin it to current latest release ie. robotframework==4.1

robocorp-code/bin/create_env/robot.yaml
avaissi: Yes, this is a quite fresh feature in RCC that it will allow multiple *_freeze.yaml files, which get prioritized from the top until a suitable platform match is found. In case no platform matches RCC uses the given condaConfigFile.

Just a nice way to wrap multi platform dependencies into one robot.yaml

robocorp-code/vscode-client/src/inspector.ts
avaissi: ah, yes. Forgot to cleanup the "old" inspector environment management out of the way. Good catch!

robocorp-code/vscode-client/src/inspector.ts
avaissi: For this I was thinking that is there a possibility for cyclic dependencies and decided to go with the commands.executeCommand()-route, but it makes sense to use internals inside one extension

robocorp-code/vscode-client/src/inspector.ts
avaissi: Removed

robocorp-code/vscode-client/src/inspector.ts
avaissi: Removed

robocorp-code/vscode-client/src/inspector.ts
avaissi: Removed

robocorp-code/vscode-client/src/inspector.ts
avaissi: Removed

robocorp-code/vscode-client/src/inspector.ts
avaissi: changed to use the internal getLanguageServerPythonInfo()

avaissi (commented):
Now all open issues should have been addressed.
Could you @fabioz retry with the latest?
There was a blatant bug in how I was calling the python -m inspector.cli, that the subprocess call didn't set the CWD for the process.. thus locators.json and other saves failed as paths were wonky

avaissi:
I'll try to get the image shown when hovering the path in locators.json fixed still for this PR

avaissi:
> After creating a new `locators.json`, it seems to work, but then, when hovering over the screenshot field, the contents of the screenshot are no longer properly shown (this is done in `robocorp_code.robocorp_language_server.RobocorpLanguageServer.m_text_document__hover` -- it actually does seem it has support for relative paths, but then, it's not working for me with this version, but I haven't investigated why.
> 
> As a note, it seems that the locators.json no longer has support for the screenshot as a base64 image (that should probably be backward compatible to be able to load it, so that we don't break existing locators -- which is probably the case for existing users of vscode).

This should now also be fixed. Problem was that reading the relative path was blocked by the browser as not permitted. Changed this instead so, that the backend reads the file and returns the content as a base64 string

PR #437, opened 2021-08-30
Refactor locators.json handling to use common loader

avaissi:
- _get_locators_json_info() and _remove_locator_from_json_internal() to use shared loader for LocatorsDatabase

### DIFF
--- a/robocorp-code/bin/create_env/conda.yaml
+++ b/robocorp-code/bin/create_env/conda.yaml
@@ -1,10 +1,9 @@
 channels:
-  - defaults
   - conda-forge
 dependencies:
-  - python=3.8
+  - python=3.8.10
   - pip=20.1
   - pip:
-    - robocorp-inspector==0.3.5
+    - robocorp-inspector==0.3.6
     - robotframework==4.1
 
--- a/robocorp-code/src/robocorp_code/robocorp_language_server.py
+++ b/robocorp-code/src/robocorp_code/robocorp_language_server.py
@@ -25,6 +25,7 @@ from robocorp_code.protocols import (
     ActionResultDictLocalRobotMetadata,
     ActionResultDictRobotLaunch,
     TypedDict,
+    ActionResultDictLocatorsJson,
     ActionResultDictLocatorsJsonInfo,
     LocatorEntryInfoDict,
     ConfigurationDiagnosticsDict,
@@ -36,7 +37,6 @@ from robocorp_ls_core.python_ls import PythonLanguageServer
 from robocorp_ls_core.robotframework_log import get_logger
 from robocorp_ls_core.watchdog_wrapper import IFSObserver
 from robocorp_ls_core import watchdog_wrapper
-
 
 log = get_logger(__name__)
 
@@ -107,7 +107,6 @@ class ListWorkspaceCachedInfoDict(TypedDict):
 
 
 class RobocorpLanguageServer(PythonLanguageServer):
-
     # V2: save the account info along to validate user.
     CLOUD_LIST_WORKSPACE_CACHE_KEY = "CLOUD_LIST_WORKSPACE_CACHE_V2"
     PACKAGE_ACCESS_LRU_CACHE_KEY = "PACKAGE_ACCESS_LRU_CACHE"
@@ -575,7 +574,6 @@ class RobocorpLanguageServer(PythonLanguageServer):
 
                 workspace_activities = activities_result.result
                 for activity_package in workspace_activities:
-
                     key = (ws.workspace_id, activity_package.robot_id)
                     sort_key = "%05d%s" % (
                         ws_id_and_pack_id_to_lru_index.get(key, DEFAULT_SORT_KEY),
@@ -775,7 +773,6 @@ class RobocorpLanguageServer(PythonLanguageServer):
         with progress_context(
             self._endpoint, "Uploading to existing robot", self._dir_cache
         ):
-
             result = self._rcc.cloud_set_robot_contents(
                 directory, workspace_id, robot_id
             )
@@ -975,11 +972,39 @@ class RobocorpLanguageServer(PythonLanguageServer):
 
         return 0, 0  # I.e.: unable to find
 
+    @staticmethod
+    def _load_locators_db(robot_yaml_path) -> ActionResultDictLocatorsJson:
+        from RPA.core.locators.database import LocatorsDatabase
+
+        locators_json = robot_yaml_path.parent / "locators.json"
+        if locators_json.exists():
+            db = LocatorsDatabase(str(locators_json))
+            db.load()
+            if not db.locators:
+                error = db.error
+                if not isinstance(error, str):
+                    if isinstance(error, tuple) and len(error) == 2:
+                        try:
+                            error = error[0] % error[1]
+                        except:
+                            error = str(error)
+                    else:
+                        error = str(error)
+
+                return {"success": False, "message": error, "result": None}
+        else:
+            return {
+                "success": False,
+                "message": f"Locators.json file does not exist: {str(locators_json)}",
+                "result": None,
+            }
+
+        return {"success": True, "message": None, "result": (db, locators_json)}
+
     @command_dispatcher(commands.ROBOCORP_GET_LOCATORS_JSON_INFO)
     def _get_locators_json_info(
         self, params: dict = None
     ) -> ActionResultDictLocatorsJsonInfo:
-        from RPA.core.locators.database import LocatorsDatabase
         from RPA.core.locators.containers import Locator
 
         if not params or "robotYaml" not in params:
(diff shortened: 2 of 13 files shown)


robocorp-code/vscode-client/src/views.ts
```diff
@@ -229,13 +229,7 @@ export class LocatorsTreeDataProvider implements vscode.TreeDataProvider<Locator
...
         if (!actionResult['success']) {
             this.lastRobotEntry = undefined;
-            return [{
-                name: actionResult.message,
-                type: "error",
-                line: 0,
-                column: 0,
-                filePath: robotEntry.robot.filePath,
-            }];
+            return [];
```
fabioz, line 232: Why this change? 
-- i.e.: if there's some error loading the info, I think it's reasonable to show it to users...

robocorp-code/src/robocorp_code/robocorp_language_server.py
```diff
@@ -990,55 +1015,41 @@ def _get_locators_json_info(
...
         locators_json_info: List[LocatorEntryInfoDict] = []
         locator: Locator
-        if locators_json.exists():
+        db, locators_json = None, None
+        try:
+            action_result: ActionResultDictLocatorsJson = self._load_locators_db(path)
+            if action_result["success"]:
+                db, locators_json = action_result["result"]
+            else:
+                raise Exception(action_result["message"])
```
fabioz, line 1026: This seems a bit odd for me... 

i.e.: Since `ActionResultDictLocatorsJsonInfo` has a value indicating the success, I think you should just return one with a `success: False` (otherwise it could just return a `List[LocatorEntryInfoDict]` and throw exceptions).

robocorp-code/src/robocorp_code/robocorp_language_server.py
```diff
@@ -990,55 +1015,41 @@ def _get_locators_json_info(
...
+                    {
+                        "name": name,
+                        "line": line,
+                        "column": col,
+                        "type": as_dict["type"],
+                        "filePath": str(locators_json),
+                    }
+                )
+        except Exception as e:
+            return {"success": False, "message": str(e), "result": None}
```
fabioz, line 1045: Exceptions should be logged (i.e.: `log.exception("Error message")`).

fabioz (changes requested):

robocorp-code/vscode-client/src/views.ts
avaissi: This populates the Locators view with unnecessary tree items, even in "normal" case of not having a locators.json for the robot, or an empty locators.json. Might be better to show errors as notifications maybe?

fabioz: I'd say that if we want to hide and count that as "normal", the language server should return `success:true` with an empty list (so that errors are still shown) and still show errors here.

avaissi: That sounds fair enough, I'll modify those cases so that the language server to respond with success and an empty list

avaissi: Handling now the empty or missing locators.json in language server and showing any errors to user as before

robocorp-code/src/robocorp_code/robocorp_language_server.py
avaissi: Modified the if-else to return rather than raise Exception

robocorp-code/src/robocorp_code/robocorp_language_server.py
avaissi: Added log.exception()

robocorp-code/bin/create_env/conda.yaml
```diff
@@ -1,10 +1,9 @@
 channels:
-  - defaults
   - conda-forge
 dependencies:
-  - python=3.8
+  - python=3.8.10
```
avaissi, line 4: Pinned the python version to avoid faulty version greeping

robocorp-code/src/robocorp_code/robocorp_language_server.py
```diff
@@ -1048,28 +1052,18 @@ def _remove_locator_from_json_internal(
...
-                            try:
-                                error = error[0] % error[1]
-                            except:
-                                error = str(error)
-                        else:
-                            error = str(error)
-
-                    return {"success": False, "message": error, "result": None}
+        db, locators_json = None, None
+        action_result: ActionResultDictLocatorsJson = self._load_locators_db(path)
```
fabioz, line 1056: I think that this line should be inside the `try..except` below as it can raise an exception when loading the db.

robocorp-code/src/robocorp_code/robocorp_language_server.py
```diff
@@ -1048,28 +1052,18 @@ def _remove_locator_from_json_internal(
...
+        try:
+            if action_result["success"]:
+                db, locators_json = action_result["result"]
+            else:
+                raise Exception(action_result["message"])
```
fabioz, line 1061: Instead of raising the exception it could be just:

return {"success": False, "message": action_result["message"], "result": None}

robocorp-code/vscode-client/src/views.ts
```diff
@@ -1,12 +1,27 @@
-import { TREE_VIEW_ROBOCORP_CLOUD_TREE, TREE_VIEW_ROBOCORP_LOCATORS_TREE, TREE_VIEW_ROBOCORP_ROBOTS_TREE, TREE_VIEW_ROBOCORP_ROBOT_CONTENT_TREE } from './robocorpViews';
+import {
```
fabioz, line 1: Note: we should probably settle into some formatter for the future so that imports aren't all reformatted all the time -- no need to change anything right now, just a reminder to myself while notifying you too ;)

robocorp-code/vscode-client/src/locators.ts
```diff
@@ -8,25 +13,41 @@ import { OUTPUT_CHANNEL } from './channel';
...
     let locatorSelected: LocatorEntry | undefined = locator || getSelectedLocator();
     if (locatorSelected) {
-        vscode.env.clipboard.writeText(locatorSelected.name);
+        env.clipboard.writeText(locatorSelected.name);
     }
 }
 
 export async function removeLocator(locator?: LocatorEntry) {
+    // Confirmation dialog button texts
+    const DELETE = 'Delete';
```
fabioz, line 22: Instead of `getSelectedLocator()`, shouldn't it use the one received as a parameter (if available)?

robocorp-code/vscode-client/src/locators.ts
```diff
@@ -8,25 +13,41 @@ import { OUTPUT_CHANNEL } from './channel';
 export async function copySelectedToClipboard(locator?: LocatorEntry) {
     let locatorSelected: LocatorEntry | undefined = locator || getSelectedLocator();
     if (locatorSelected) {
-        vscode.env.clipboard.writeText(locatorSelected.name);
+        env.clipboard.writeText(locatorSelected.name);
```
fabioz, line 16: Instead of `getSelectedLocator()`, shouldn't it use the one received as a parameter (if available)?

fabioz (changes requested):

robocorp-code/vscode-client/src/views.ts
avaissi: Right, that makes sense. Maybe we should config a prettier/linter configs for python and typescript/javascript as well, if not already there.. I've tried to remember to format python with black at least :)

robocorp-code/vscode-client/src/locators.ts
avaissi: let locatorSelected: LocatorEntry | undefined = locator || getSelectedLocator();
The || notation should take care of prioritizing the parameter value.. could be ?? as well to strictly compare it with undefined (optional value)

robocorp-code/vscode-client/src/locators.ts
avaissi: This might be just a bit poor choice of variable name for the `locatorSelected` . The previous comment applies here as well

fabioz: You're right, sorry, my bad upon reviewing.

robocorp-code/vscode-client/src/locators.ts
fabioz: You're right, sorry, my bad upon reviewing.

robocorp-code/src/robocorp_code/robocorp_language_server.py
avaissi: Damn, missed this with the `raise Exception` changes to returns.. fixing

robocorp-code/src/robocorp_code/robocorp_language_server.py
avaissi: added inside try..except

PR #454, opened 2021-09-27
Matti/work items cont

mmokko:


robocorp-code/vscode-client/src/viewsRobotSelection.ts
```diff
@@ -0,0 +1,140 @@
...
+  robotSelectionChanged(robotEntry: RobotEntry) {
+      // When the robot selection changes, we need to start tracking file-changes at the proper place.
+      if (this.lastWatcher) {
+          this.lastWatcher.dispose();
+          this.lastWatcher = undefined;
+      }
+      this.fireRootChange();
+
+      let d = basename(dirname(robotEntry.uri.fsPath));
+      let watcher: vscode.FileSystemWatcher = vscode.workspace.createFileSystemWatcher('**/' + d + '/**', false, true, false);
```
fabioz, line 65: I know this is this way since it was already like that due to `RobotContentTreeDataProvider`, but after reviewing this API again, I think it should be something as:

`vscode.workspace.createFileSystemWatcher(new vscode.RelativePattern(robotEntry.uri, '**/**'), ...`

At this place we could potentially deal only with changes in the `devdata` folder for the case where we're dealing with work items (so, it should probably be customizable).

robocorp-code/vscode-client/src/activities.ts
```diff
@@ -664,6 +664,7 @@ export async function createRobot() {
 export async function updateLaunchEnvironment(args) {
     let robot = args['targetRobot'];
     let environment = args['env'];
+    let task = args['task'] || 'latest-run';
```
fabioz, line 667: I think this may be confusing to the user if sometimes it's saved in one place and other times in another (so, I think that either we should always have a `task` or always use `latest-run` -- or use the previous approach for numbering runs).

robocorp-code/vscode-client/src/viewsWorkItems.ts
```diff
@@ -0,0 +1,203 @@
...
+  return filename;
+}
+
+async function createNewWorkItem(workItemInfo: WorkItemsInfo, workItemName: string): Promise<void> {
+  if (!workItemInfo?.input_folder_path || !workItemName) {
+      return;
+  }
+
+  const targetFolder = join(workItemInfo.input_folder_path, workItemName);
+  const targetFile = join(targetFolder, 'work-items.json');
```
fabioz, line 47: Should it check if the `targetFile` already exists before overwriting?

robocorp-code/vscode-client/src/viewsWorkItems.ts
```diff
@@ -0,0 +1,203 @@
...
+export async function newWorkItemInWorkItemsTree(): Promise<void> {
+  const workItemInfo = await getWorkItemInfo();
+  const workItemName = await queryNewWorkItemName();
+  await createNewWorkItem(workItemInfo, workItemName);
+}
+
+export async function deleteWorkItemInWorkItemsTree(): Promise<void> {
+  let robotContentTree = treeViewIdToTreeView.get(TREE_VIEW_ROBOCORP_WORK_ITEMS_TREE);
+  if (!robotContentTree) {
+      return undefined;
```
fabioz, line 67: `await vscode.window.showInformationMessage("No Robot selected for work item selected for deletion.")`

robocorp-code/vscode-client/src/viewsWorkItems.ts
```diff
@@ -0,0 +1,203 @@
...
+         "message": "Hello World!"
+      },
+      "files": {
+          "orders.xlsx": "orders.xlsx"
+      }
+  }
+]`;
+
+async function getWorkItemInfo(): Promise<WorkItemsInfo | null> {
+  // Would there be a faster way of getting the work item path other than querying the work item info?
```
fabioz, line 22: Well, since we're dealing with the tree and the related command is hidden from the command pallette, you could get it  (with `treeViewIdToTreeView.get(TREE_VIEW_ROBOCORP_WORK_ITEMS_TREE);` and then have some getter for:

`WorkItemsTreeDataProvider.workItemsInfo`.

It does make more sense to just use the info already cached in the tree in this case.

Although note that if we decided to show the command in the command palette later on, the code in this method would be needed as a fallback, so, we may want to keep it in the case where the tree wouldn't be available...

robocorp-code/vscode-client/src/viewsWorkItems.ts
```diff
@@ -0,0 +1,203 @@
...
+  // The work item tree provider does have it already, but couldn't access it through existing interface.
+  // Would also prefer the keep the static path definition only at the server side.
+  const currTreeDir: FSEntry | undefined = await getCurrRobotDir();
```
fabioz, line 25: Should handle the case where `currTreeDir` is undefined.

fabioz (changes requested):

robocorp-code/vscode-client/src/activities.ts
mmokko: I added the latest-run only as a fallback if for some reason the task value is not provided. If you think it's better to make it a mandatory argument and throw an error in case task name is not provided, I think that would a good solution as well.

I asked about the numbered run vs task name as output in the demo and the design preferred the task name approach, I think it originally came from Janne anyway. I demoed that the next run with same task will overwrite the work item data and it was accepted.

fabioz: Ok, seems fair to make the task name mandatory then.

fabioz: As a note, there are cases where the task name is not available (when the run is made from the language server and not robocorp code) and using the task name is also dangerous (we need to escape characters because not all task names are valid filesystem entries depending on the OS).

So, I just left it as `latest-run`.

As we won't be cycling names, the work items library needs to be careful when it's rewriting the entries and dealing with locks on the open files for that on the filesystem level (but that'd be the same case even with the task name, just with a lower chance of happening).

PR #460, opened 2021-09-30
Disable Work Items feature for releasing the latest RCC

mmokko:
Disable the Work Item feature.

While testing the version where work items were disabled, I received an error from the lsp that the environ parameter was called for  a null object. Fixed a couple of places where that could happen and that fixed the issue.

Also noticed that the codegen included the package.json scripts, so the earlier prettier scripts were wiped when codegen was run. Added them to the proper places now.

Had some trouble getting black linter to pass in the pipeline. It turned out that I had to have the exact version "pip install black==21.9b0" to make it pass. I had a newer release and it does a different format that the old version doesn't pass. Maybe this should be added to the test_requirements.txt as well?


robocorp-code/codegen/commands.py
```diff
@@ -384,30 +384,30 @@ def __init__(
...
-    ),
-    Command(
-        "robocorp.helpWorkItems",
-        "Work Items Help",
-        add_to_package_json=True,
-        server_handled=False,
-        icon="$(question)",
-        hide_from_command_palette=True,
-    ),
+    # Command(
```
fabioz, line 387: I was actually thinking about something a bit less intrusive to disable it.

For the commands, there's already an `add_to_package_json` which can be set to `False` in the `Command` constructor (in `codegen/commands.py`), so, you don't need to comment it out entirely, and I was thinking about doing the same for the `TreeViewContainer` (and then fixing the code generation that adds that and just don't add it to the `package.json` -- which should be simple as it'd be just an additional check when iterating `TREE_VIEW_CONTAINERS` on the few places that use it). 

Aside from that, the only other thing missing would be not binding `updateLaunchEnvironment` in the `extension.js` and not binding the tree in `views.ts` (and I believe all the remainder of the code can be kept as is since constants and definitions would be there -- even though there'd be no way for the user to activate any of the related commands/tree).

Also, it'd be nice separate commits adding those features /making the interpreter null fixes and afterwards a single one that disables things afterwards so that it's just a matter of reverting it to re-enable the features (right now the first commit mixes prettier with disabling).

Does that seem reasonable?

fabioz (changes requested):

robocorp-code/codegen/commands.py
fabioz: Or maybe we can do that single commit to disable the features in a branch and create a release from the branch (that way we don't need to change the master just for this release -- in which case it may be reasonable to not do the improvements to add to `package.json` if you feel it's more work than it should be).

Other fixes (such as the interpreter null check as well as prettier) should be in master anyways though.

fabioz:
Closing this PR since other pull requests are being created to address this.

PR #466, opened 2021-10-04
Save last 5 work items outputs with output-prefix

mmokko:
Fix an issue where the next work item output name was incorrectly
incremented twice. This occured with two consecutive runs as the
algorithm was saving the next output name from the previous run instead
of previous one.

Make sure last 5 runs are always kept, even if there are outputs with
different names. Previously only last 4 runs were saved and all other
named output work items were counted as well.


robocorp-code/src/robocorp_code/robocorp_language_server.py
```diff
@@ -825,7 +822,7 @@ def _compute_new_output_workitem_path(
 
         next_run = max_run + 1
         self._prefix_to_last_run_number_and_time[output_prefix] = (
-            next_run,
```
fabioz, line 828: This isn't a real fix as 2 runs may end up running with the same id if a run is done before the next run creates the folder related to the output.

Rather, the proper fix is changing this function to accept a parameter for a case where the output shouldn't be incremented (the issue is that there are calls done to this function which shouldn't increment this number).

fabioz (changes requested):

fabioz:
Closing PR (the fix was integrated in: https://github.com/robocorp/robotframework-lsp/commit/135a302fa83041e118471945a1f1477b54f34141).

PR #470, opened 2021-10-07
Matti/dynamic robot templates

mmokko:
Update RCC to v11.3.2.

Get the template selection for creating a new robot from RCC.
RCC fetches the latest templates from remote repository, enabling dynamic updating for templates. And possibility to add more templates dynamically.

### DIFF
--- a/robocorp-code/src/robocorp_code/rcc.py
+++ b/robocorp-code/src/robocorp_code/rcc.py
@@ -67,7 +67,7 @@ def download_rcc(location: str, force: bool = False) -> None:
                     else:
                         relative_path = "/linux32/rcc"
 
-                RCC_VERSION = "v11.2.0"
+                RCC_VERSION = "v11.3.2"
                 prefix = f"https://downloads.robocorp.com/rcc/releases/{RCC_VERSION}"
                 url = prefix + relative_path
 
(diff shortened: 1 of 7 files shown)


robocorp-code/src/robocorp_code/rcc.py
```diff
@@ -324,33 +326,25 @@ def _run_rcc(
...
         if not result.success:
             return ActionResult(success=False, message=result.message)
 
-        output = result.result
+        output: Dict[str, str] = []
+        try:
+            output = json.loads(result.result)
+        except json.decoder.JSONDecodeError as e:
+            return ActionResult(success=False, message=f"Invalid output format: {e}")
         if output is None:
```
fabioz, line 340: Instead of checking `output is None`, this should check if `result.result is None` (this should be done before trying to load the json from it).

fabioz (changes requested):
Nice work. Just one minor issue...

robocorp-code/src/robocorp_code/rcc.py
mmokko: Good catch! I pushed the fix in a separate commit, let me know if you want me to squash the commits, if you are not merging with squash.

Issue #397, opened 2021-07-13
Load libraries more than once but with different alias


jkRobotic:
In robot Framework, it is possible to load a library more than once, but with different alias and parameters.

This should work:
```
Library    Collections    WITH NAME    col1
Library    Collections    WITH NAME    col2
```
According to the example above, only the second import (col2) is recognized and the VSCode marked the first import (col1) as **"Undefined Keyword"** (see picture)

![image](https://user-images.githubusercontent.com/87384643/125492485-602ee3ac-a4cd-45c2-9ecc-899658bc5044.png)

**Should work in this way:**
In this case **"col1"** and **"col2"** alias should access to the same keyword list from **"Collections"** library, as described on robotframework documentation page (http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#using-library-name)

![image](https://user-images.githubusercontent.com/87384643/125490749-f9842853-258f-4ace-8769-73591686cdb6.png)

This feature would be a big benefit if more than one connections should use in the same testcase. (like the example on the picture)

Thanks in advance!

PR #494, opened 2021-11-07
robotframework_ls: handle remote libraries import

### LINKED ISSUE #343: Remote libraries with URL other than http://127.0.0.1:8270 are not supported
**Describe the bug**
If you run a remote library on the default URL `http://127.0.0.1:8270` syntax highlighting, code completion and hovering (quick documentation) will work. This is the default address that the Robot Framework Remote library uses if no address is given, see http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#importing-remote-library.

If you deviate from this default, that functionality wil not work. For example when you run your remote library from the following URLs:
```
http://127.0.0.2:8270
http://127.0.0.1:8280
http://127.0.0.1:8270/path
```
When you run or have run a remote library at `http://127.0.0.1:8270` it will show the keywords of that library for another remote library running on a deviating URL. 

When you have run a remote library at `http://127.0.0.1:8270` and you run another remote library at the same URL, it wil still show the keywords of the first remote library.

**To Reproduce**
***General preparation***
1. Plugin _Robot Framework Language Server_ is installed and enabled in PyCharm
1. Close PyCharm
2. Install Python Remote Server for Robot Framework
`pip install robotremoteserver`
3. Remove stored remote library keywords, if any:
`del <user dir>\.robotframework-ls\specs\v2\<????????>_4.0.1\user\Remote.libspec*`
4. Unzip demo files
[Demo.zip](https://github.com/robocorp/robotframework-lsp/files/6445936/Demo.zip)

***Demo 1: no recognition keywords from remote libraries on not default addresses***
1. Open a command prompt in demo directory _remote-libs_ and start de remote Collections library
`py remotecollections.py 127.0.0.1 8280`
2. Open a second command prompt in demo directory _remote-libs_ and start de remote String library
`py remotestring.py 127.0.0.2 8270`
3. Open a third command prompt and verify that both remote libraries can provide a list of keywords
`py -m robot.libdoc Remote::http://127.0.0.1:8280 list`
`py -m robot.libdoc Remote::http://127.0.0.2:8270 list`
4. Open PyCharm
5. Open demo directory _remote-libs_
6. Open _test_remote_collections.robot_ and _test_remote_string.robot_
7. Verify that PyCharm gives the error undefined keyword for a correct remote library keyword
8. Verify that there are no autocompletion suggestions if you type `RemCols.` or `RemStr.` respectively
9. Verify that no file  _Remote.libspec_ has been generated in directory _<user dir>\.robotframework-ls\specs\v2\<????????>_4.0.1\user_

***Demo 2: suggesting keywords from remote library running at default address***
1. Close PyCharm
2. Open a fourth command prompt in demo directory _remote-example_ and start de remote example library
`py examplelibrary.py`
3. In the third command prompt window verify that the remote example library can provide a list of keywords
`py -m robot.libdoc Remote::http://127.0.0.1:8270 list`
4. Open PyCharm again
5. Verify that PyCharm now suggests keywords from the example library if you type `RemCols.` or `RemStr.` respectively
6. Verify that a file  _Remote.libspec_ has been generated in directory _<user dir>\.robotframework-ls\specs\v2\<????????>_4.0.1\user_

***Demo 3: Correct behavior for remote library on default address***
1. In PyCharm open demo directory _remote-example_
2. Open _tests.robot_
3. Verify that syntax highlighting, code completion and hovering (quick documentation) do work for the keywords provided by the remote example library

***Demo 4: Same behavior after stopping remote library on default address***
1. Close PyCharm
2. Stop the remote example library by pressing Ctrl+C in the fourth command prompt window
3. Repeat demo 2 (from step 4) and demo 3 and verify behavior is still the same.

***Demo 5: Same behavior after running another remote library address***
1. Close PyCharm
2. Stop the remote Collections library by pressing Ctrl+C in the first command prompt window
3. Start the remote Collection library on the default address
`py remotecollections.py`
4. In the third command prompt window verify that the remote collections library now provide a list of keywords at the default address
`py -m robot.libdoc Remote::http://127.0.0.1:8270 list`
5. Repeat demo 2 (from step 4) and demo 3 and verify behavior is still the same (i.e. the keywords from the remote example library have not been replaced by the keywords of the remote collections library)

***No demo for address with path***
Unfortunately I do not know how to create a demo remote library that uses a path in the address, e.g. `http://127.0.0.1:8270/path`. Based on my work with a remote library that I am not able to share I can confirm that behavior for a remote library on an address like `http://127.0.0.1:8270/path` is the same as for the other remote libraries not running on the default address.

**Expected behavior**
1. Keyword information for a remote library is collected from the remote library at the address as specified in its Library statement under Settings, taking into account ip address, port and path
2. Keyword information is removed if the remote library stopped running at the indicated address
3. Keyword information is replaced if the remote library is switched for a certain address

I think 2 and 3 are not that important since typically you run a remote library always on the same address. I think it is an acceptable workaround to delete Remote.libspec if you need to change or upgrade a remote library on a certain address.

**Versions:**
 - OS: Windows 10
 - Robot Framework Version 4.0.1
 - Robot Framework Language Server Version 0.15.0
 - Client Version PyCharm Community Edition 2021.1.1

**Logs**
I do not think logging is necessary in this case. If you need it let me know for which situation and which purpose.

### LINKED ISSUE #373: Consider library arguments when creating libspec (Keywords may be available only given some arguments)
**Describe the bug**
When import FakerLibrary with additional argument in this case " locale=PL" additional Keywords should be visible ex. "Pesel"

**To Reproduce**
1.  Import "Library           FakerLibrary    locale=PL"
2.  Search for keywords "Pesel"

Example Code
[Bug.txt](https://github.com/robocorp/robotframework-lsp/files/6554559/Bug.txt)

**Expected behavior**
Additional Keywords should be visible

**Screenshots**
image

image

**Versions:**
 - OS: [ Windows 10]
 - Robot Framework Version  4.0.2
 - Robot Framework Language Server Version .17.0
 - Client Version  VSCode 1.56.2

**Logs**
image

JoeyGrajciar:
Now the libspec generation will be called with lib in format `<Library>::<arg0>::...::<argN>` to match what
LibDoc expects.

Fixes #343
Fixes #373 

Obligatory gif from Neovim:
nvim_robot_lsp_remote_lib

### DIFF
--- a/robotframework-ls/src/robotframework_ls/impl/ast_utils.py
+++ b/robotframework-ls/src/robotframework_ls/impl/ast_utils.py
@@ -667,3 +667,7 @@ def create_range_from_token(token):
     }
     code_lens_range: RangeTypedDict = {"start": start, "end": end}
     return code_lens_range
+
+
+def get_library_arguments_serialized(library) -> Optional[str]:
+    return "::".join(library.args) if library.args else None
--- a/robotframework-ls/src/robotframework_ls/impl/ast_utils.py
+++ b/robotframework-ls/src/robotframework_ls/impl/ast_utils.py
@@ -540,20 +540,6 @@ def get_keyword_name_token(ast, token):
     return None
 
 
-def is_remote_library_node(node) -> bool:
-    return node.name and node.name.lower() == "remote"
-
-
-def get_library_name_from_node(node) -> Optional[str]:
-    if not is_remote_library_node(node):
-        return node.name
-
-    if len(node.args) == 0:
-        return None
-
-    return "Remote::" + node.args[0]
-
-
 def get_library_import_name_token(ast, token):
     """
     If the given ast node is a library import and the token is its name, return
@@ -565,11 +551,7 @@ def get_library_import_name_token(ast, token):
         and isinstance_name(ast, "LibraryImport")
         and ast.name == token.value  # I.e.: match the name, not the alias.
     ):
-        token.value = get_library_name_from_node(ast)
-        if (
-            token.value
-        ):  # check if it was not `Library    Remote` without actual arguments
-            return token
+        return token
     return None
 
 
--- a/robotframework-ls/src/robotframework_ls/impl/ast_utils.py
+++ b/robotframework-ls/src/robotframework_ls/impl/ast_utils.py
@@ -540,6 +540,20 @@ def get_keyword_name_token(ast, token):
     return None
 
 
+def is_remote_library_node(node) -> bool:
+    return node.name and node.name.lower() == "remote"
+
+
+def get_library_name_from_node(node) -> Optional[str]:
+    if not is_remote_library_node(node):
+        return node.name
+
+    if len(node.args) == 0:
+        return None
+
+    return "Remote::" + node.args[0]
+
+
 def get_library_import_name_token(ast, token):
     """
     If the given ast node is a library import and the token is its name, return
@@ -551,7 +565,11 @@ def get_library_import_name_token(ast, token):
         and isinstance_name(ast, "LibraryImport")
         and ast.name == token.value  # I.e.: match the name, not the alias.
     ):
-        return token
+        token.value = get_library_name_from_node(ast)
+        if (
+            token.value
+        ):  # check if it was not `Library    Remote` without actual arguments
+            return token
     return None
 
 
--- a/robotframework-ls/src/robotframework_ls/impl/collect_keywords.py
+++ b/robotframework-ls/src/robotframework_ls/impl/collect_keywords.py
@@ -277,7 +277,7 @@ def _collect_libraries_keywords(
             (
                 library.name,
                 library.alias,
-                "::".join(library.args),
+                ast_utils.get_library_arguments_serialized(library),
             )
             for library in libraries
         )
--- a/robotframework-ls/src/robotframework_ls/impl/collect_keywords.py
+++ b/robotframework-ls/src/robotframework_ls/impl/collect_keywords.py
@@ -250,7 +250,7 @@ def _collect_current_doc_keywords(
     _collect_completions_from_ast(ast, completion_context, collector)
 
 
-_LibInfo = namedtuple("_LibInfo", "name, alias, builtin, remote")
+_LibInfo = namedtuple("_LibInfo", "name, alias, builtin, args")
 
 
 def _collect_libraries_keywords(
@@ -271,19 +271,19 @@ def _collect_libraries_keywords(
             completion_context.token_value_resolving_variables(name),
             alias,
             False,
-            remote,
+            args,
         )
-        for name, alias, remote in (
+        for name, alias, args in (
             (
-                ast_utils.get_library_name_from_node(library),
+                library.name,
                 library.alias,
-                ast_utils.is_remote_library_node(library),
+                "::".join(library.args),
             )
             for library in libraries
         )
         if name
     )
-    library_infos.add(_LibInfo(BUILTIN_LIB, None, True, False))
+    library_infos.add(_LibInfo(BUILTIN_LIB, None, True, None))
     libspec_manager = completion_context.workspace.libspec_manager
 
     for library_info in library_infos:
@@ -296,7 +296,7 @@ def _collect_libraries_keywords(
             create=True,
             current_doc_uri=completion_context.doc.uri,
             builtin=library_info.builtin,
-            remote=library_info.remote,
+            args=library_info.args,
         )
         if library_doc is not None:
             #: :type keyword: KeywordDoc
--- a/robotframework-ls/src/robotframework_ls/impl/collect_keywords.py
+++ b/robotframework-ls/src/robotframework_ls/impl/collect_keywords.py
@@ -250,7 +250,7 @@ def _collect_current_doc_keywords(
     _collect_completions_from_ast(ast, completion_context, collector)
 
 
-_LibInfo = namedtuple("_LibInfo", "name, alias, builtin")
+_LibInfo = namedtuple("_LibInfo", "name, alias, builtin, remote")
 
 
 def _collect_libraries_keywords(
@@ -262,18 +262,28 @@ def _collect_libraries_keywords(
     # Get keywords from libraries
     from robotframework_ls.impl.robot_constants import BUILTIN_LIB
     from robocorp_ls_core.lsp import CompletionItemKind
+    from robotframework_ls.impl import ast_utils
 
     libraries = completion_context.get_imported_libraries()
 
     library_infos = set(
         _LibInfo(
-            completion_context.token_value_resolving_variables(library.name),
-            library.alias,
+            completion_context.token_value_resolving_variables(name),
+            alias,
             False,
+            remote,
         )
-        for library in libraries
+        for name, alias, remote in (
+            (
+                ast_utils.get_library_name_from_node(library),
+                library.alias,
+                ast_utils.is_remote_library_node(library),
+            )
+            for library in libraries
+        )
+        if name
     )
-    library_infos.add(_LibInfo(BUILTIN_LIB, None, True))
+    library_infos.add(_LibInfo(BUILTIN_LIB, None, True, False))
     libspec_manager = completion_context.workspace.libspec_manager
 
     for library_info in library_infos:
@@ -286,6 +296,7 @@ def _collect_libraries_keywords(
             create=True,
             current_doc_uri=completion_context.doc.uri,
             builtin=library_info.builtin,
+            remote=library_info.remote,
         )
         if library_doc is not None:
             #: :type keyword: KeywordDoc
--- a/robotframework-ls/src/robotframework_ls/impl/find_definition.py
+++ b/robotframework-ls/src/robotframework_ls/impl/find_definition.py
@@ -237,6 +237,7 @@ def find_definition(completion_context: ICompletionContext) -> Sequence[IDefinit
                 completion_context.token_value_resolving_variables(token),
                 create=True,
                 current_doc_uri=completion_context.doc.uri,
+                remote=ast_utils.is_remote_library_node(token_info.node),
             )
             if library_doc is not None:
                 definition = _DefinitionFromLibrary(library_doc)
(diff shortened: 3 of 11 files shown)


fabioz:
@JoeyGrajciar thank you very much for the pull request.

Do you think it'd be possible for you to provide a test for this feature?

The main work at that would probably be having a remote library running locally during the test run to test it (so, the test must setup the remote library to be running for it to be tested). 

You can see related tests at:

https://github.com/robocorp/robotframework-lsp/blob/master/robotframework-ls/tests/robotframework_ls_tests/completions/test_keyword_completions.py

So, it'd be a matter of extending that module to create an additional test to check for this specific situation.

Also, the pull request broke existing tests (which should be fixed so that this can be integrated).

p.s.: For instructions on creating an environment to run the tests, you can see: https://github.com/robocorp/robotframework-lsp/blob/master/robocorp-code/docs/develop.md#creating-a-local-environment-for-python-development

Then, with that environment activated you can use `py.test` to run the tests locally.

JoeyGrajciar:
Sure will add tests for remote :) Shouldn't be hard using `robotremoteserver`.

.github/workflows/tests-robotframework-intellij.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       with:
         java-version: 11
     - name: Install deps
-      run: pip install --upgrade pip fire twine wheel setuptools robotframework
+      run: pip install --upgrade pip fire twine wheel setuptools robotframework robotremoteserver
```
fabioz, line 52: I don't think you're testing the `robotremoteserver` with the `Intellij` plugin (which is ok), but then, this shouldn't be needed (i.e.: this file is for the tests related to the `Robot Framework Language Server Intellij Plugin` ..

.github/workflows/tests-robotframework-interactive.yml
```diff
@@ -58,6 +58,8 @@ jobs:
     - name: Install robotframework from pip
       run: python -W ignore -m pip install "robotframework>=4.0"
       if: contains(matrix.name, '-pip-v4')
+    - name: Install robotremoteserver via pip
```
fabioz, line 61: I don't think you're testing the robotremoteserver with the interactive console (which is ok), but then, this shouldn't be needed (i.e.: this file is for the tests related to the interactive console).

robotframework-ls/src/robotframework_ls/impl/ast_utils.py
```diff
@@ -540,6 +540,20 @@ def get_keyword_name_token(ast, token):
     return None
 
 
+def is_remote_library_node(node):
+    return node.name and node.name.lower() == "remote"
+
+
+def get_library_name_from_node(node):
```
fabioz, line 547: Please add the typing info to the return (`-> Optional[str]`).

robotframework-ls/src/robotframework_ls/impl/ast_utils.py
```diff
@@ -540,6 +540,20 @@ def get_keyword_name_token(ast, token):
     return None
 
 
+def is_remote_library_node(node):
```
fabioz, line 543: Please add typing info to the return `-> bool`.

robotframework-ls/src/robotframework_ls/impl/collect_keywords.py
```diff
@@ -262,18 +262,22 @@ def _collect_libraries_keywords(
...
     from robocorp_ls_core.lsp import CompletionItemKind
+    from robotframework_ls.impl import ast_utils
 
     libraries = completion_context.get_imported_libraries()
 
     library_infos = set(
         _LibInfo(
-            completion_context.token_value_resolving_variables(library.name),
+            completion_context.token_value_resolving_variables(
+                ast_utils.get_library_name_from_node(library)
```
fabioz, line 272: I think you should check whether `ast_utils.get_library_name_from_node(library)` returns `None` here.

i.e.: While the user is typing, it's possible for him to just write a `Library   Remote`, and I'm not sure the rest of the code deals with a `None` name properly.

robotframework-ls/src/robotframework_ls/impl/find_definition.py
```diff
@@ -237,6 +237,7 @@ def find_definition(completion_context: ICompletionContext) -> Sequence[IDefinit
                 completion_context.token_value_resolving_variables(token),
                 create=True,
                 current_doc_uri=completion_context.doc.uri,
+                remote=ast_utils.is_remote_library_node(token_info.node),
```
fabioz, line 240: In this case the remote library name is not being changed (probably to get to the actual `Remote` library in a find definition).

It'd be nice to have a test for this case in: 
`robotframework-ls/tests/robotframework_ls_tests/test_find_definition.py`

fabioz (changes requested):
Thanks for the fixes!

Added some additional notes on things to be addressed for the integration...

As a note, the `Robocorp Code` tests seem to have failed due to a network hiccup, so, you can ignore those, but the tests with `Robot Framework v3` seem to be actually failing...

If the integration is not supposed to run with `Robot Framework v3`, please make a skip for that case (otherwise the test should be fixed).

robotframework-ls/src/robotframework_ls/impl/find_definition.py
JoeyGrajciar: It actually is :) It is just not in the diff of this file. `token` is acquired by calling `ast_utils.get_library_import_name_token` that returns what we need further.

fabioz: Humm, ok, just saw that... I'm not sure that's ideal. 

What happens in the case where `token.value = get_library_name_from_node(ast)` but `get_library_name_from_node(ast)` returns `None`?

-- the main things here would be the behavior for `code analysis`/`code-completion`/`find definition`.

i.e.: the user created a library with `Remote` but still hasn't specified any arguments.

fabioz (approved):
I'm approving, but I still want to do some manual testing on my side prior to merging (which I plan to do today or at most tomorrow).

JoeyGrajciar:
Sure :) Should I autosquash those fixups before merge?

fabioz:
> Sure :) Should I autosquash those fixups before merge?

Don't worry, I'll squash them during the merge.

JoeyGrajciar:
Wait with merge i might have found different solution (probably simpler) :)

JoeyGrajciar:
+ bonus now arguments are supported :)

JoeyGrajciar:
Okay I guess that is the last change :) You can review.

fabioz:
Humm, actually, for arguments I think we don't want something like that...

In the general case, a library will always have the same API regardless of its arguments, there are only 2 exceptions I know of for this:
- Remote
- FakerLibrary (https://github.com/robocorp/robotframework-lsp/issues/373)

So, I think that we should initially just hardcode to handle the arguments for those 2 cases (in the future we can create some setting if more cases are identified).

i.e.: if there's some custom library which receives an url as a parameter, I don't think we should create potentially hundreds of libspecs in the user machine -- generating a libspec is both slow and space-wasteful.

JoeyGrajciar:
In my company we have quite a few files with one local library with args and at least two remote libraries.
Do you suggest hardcoding this args stuff only for `FakerLibrary` and `Remote` libname instead of generic solution? :)

fabioz:
If you foresee more use cases, we could do a setting from the start to note for which libraries parameters should be taken into consideration (the setting by default would have `Remote` and `FakerLibrary`) -- the only thing is that I think that by default generating a new libspec for a library whenever an argument changes isn't ideal (especially because it may be triggered while the user is typing).

So, I agree with implementing a generic solution but that feature (using args to generate the libspec and consider a new libspec whenever an argument changes) should be used only on the cases we know it'll benefit the user (which isn't the case for most libraries).

Also, if args are taken into account for the library, we probably need some testing in the auto-import to write the import properly for such cases (i.e.: a test in `robotframework-ls\tests\robotframework_ls_tests\completions\test_auto_import.py` to handle the auto-import from a library where the arguments are important to know the keywords available for such library is needed).

JoeyGrajciar:
> generating a new libspec for a library whenever an argument changes isn't ideal

Yes that is a bummer. Thing is it is kind of hard to determine whether it is the same library but with different order of named args, changed "insignificant" argument or some other things.

Some form of setting would shift the responsibility to user to tell us based on which parameter we should identify correct libspec in case of inconsistency, multiargument libraries etc.

JoeyGrajciar:
Won't this actually also implement this [Load libraries more than once but with different alias #397](https://github.com/robocorp/robotframework-lsp/issues/397) for libraries with args?

fabioz:
> Won't this actually also implement this [Load libraries more than once but with different alias #397](https://github.com/robocorp/robotframework-lsp/issues/397) for libraries with args?

I think that's a different issue (it does need this as a pre-requisite for the arguments handling for his real-world case, but the alias in the example is using the same library twice and it still fails... as it seems a different issue it should probably be solved in a different PR -- this one is already tackling 2 issues ;)

JoeyGrajciar:
Was thinking about the optimization of not generating libspecs when arguments change. Currently when I am typing library import basically each stroke attempts to generate libspec e.g. `Library    Remote    ` is written there is attempt to generate libspec for `Remote` library which goes to `127.0.0.1:8270` which may or may not be available then adding another character makes another request this one should most probably fail. I think this is not exactly issue introduced by these changes even `Library    Built` would be attempt and mayber user has library with name `Built` but finishes to `Builtin` then there would be 2 libspecs.

So I propose that this optimization should be done outside of this PR.

fabioz:
Integrated by rebasing:
https://github.com/robocorp/robotframework-lsp/commit/0783584477dd0ef72a1a1c0710b38ebcc71ebf94

and doing a subsequent commit to add settings to filter to which libraries the new heuristic should be applied (`"robot.libraries.libdoc.needsArgs"`) -- if we find a better approach afterwards we can remove the setting.
https://github.com/robocorp/robotframework-lsp/commit/23836a699e904956ba0d8f4b2e4aed40c99c13d8

It's possible to specify `"robot.libraries.libdoc.needsArgs": ["*"]` to always consider arguments important or to specify just a few libraries such as: `"robot.libraries.libdoc.needsArgs": ["remote", "SomeOtherLib"]` -- default is `remote` and `fakerlib`.

PR #500, opened 2021-11-09
fix missing colons in versions section

DetachHead:


fabioz:
Thanks for the PR. Just merged.

PR #521, opened 2021-12-10
Use the global launch template if present

mardukbp:
A global launch template can be defined in the *launch* section of `settings.json`, but currently the RF LS extension ignores it. With this PR if a global launch template is defined, it takes precedence over the launch template in launch.json, as expected. If there is no global launch template, then the template in launch.json is used.

### DIFF
--- a/robotframework-ls/vscode-client/src/run.ts
+++ b/robotframework-ls/vscode-client/src/run.ts
@@ -53,6 +53,10 @@ async function checkFileExists(file) {
 }
 
 async function readAllDebugConfigs(workspaceFolder: WorkspaceFolder): Promise<DebugConfiguration[]> {
+    const launch = workspace.getConfiguration("launch");
+    const configs = launch.inspect<DebugConfiguration[]>("configurations")?.globalValue;
+    if (configs) return configs;
+
     const filename = path.join(workspaceFolder.uri.fsPath, ".vscode", "launch.json");
     if (!(await checkFileExists(filename))) {
         return [];


fabioz:
@mardukbp thanks for the pull request!

I'd like to merge this, but it seems that this commit isn't tied to your user for some reason (so, it's not recognizing that you already signed the `Contributor License Agreement`).

So, can you take a look into fixing the issue raised by the bot (https://github.com/robocorp/robotframework-lsp/pull/521#issuecomment-991087785) so that I can merge it?

mardukbp:
@fabioz Done. Thanks for merging it so quickly!

robotframework-ls/vscode-client/src/run.ts
```diff
@@ -53,6 +53,10 @@ async function checkFileExists(file) {
 }
 
 async function readAllDebugConfigs(workspaceFolder: WorkspaceFolder): Promise<DebugConfiguration[]> {
+    const launch = workspace.getConfiguration("launch");
+    const configs = launch.inspect<DebugConfiguration[]>("configurations").globalValue;
```
fabioz, line 57: It seems that `launch.inspect<DebugConfiguration[]>("configurations")` could return `undefined`, so, maybe this should be:

`launch.inspect<DebugConfiguration[]>("configurations")?.globalValue`

mardukbp: You are right. VS Code did not warn me, because `strict` is not set in `vscode-client/tsconfig.json`. I think it should be set to true to prevent running into runtime errors.

fabioz: Agreed. I created https://github.com/robocorp/robotframework-lsp/issues/523 to track that.

fabioz:
Just merged. Thanks for the PR!

Issue #586, opened 2022-02-24
Embedded variables not highlighted in keyword calls


weltings:
**Describe the bug**
Syntax highlighting does not recognize embedded variables in BDD style testing

**To Reproduce**

*** Test Cases ***

```
Example test case
    Given A UM-MIG compliant B1 declaration with all mandatory elements
    And Licence is specified with type C068 reference NL0074CDIU0571970 amount 1 and currency EUR
    When The declaration is submitted to EU customs
    Then Message DVV02: Validate Customer Licence should have been received
    And The message should have value JA in element resultaatcode

*** Keywords ***

Licence is specified with type ${type} reference ${reference} amount ${amount} and currency ${currency}
    ${index}=    Get sequence numeric for new SupportingDocument element in section GovernmentAgencyGoodsItem
    In the section GovernmentAgencyGoodsItem an element with tag SupportingDocument has been added
    In the section SupportingDocument[${index}] the element ID has been added with value ${reference}
    In the section SupportingDocument[${index}] the element TypeCode has been added with value ${type}
    In the section SupportingDocument[${index}] the element SequenceNumeric has been added with value ${index}
    In the section SupportingDocument[${index}] an element with tag WriteOff exists
    In the section SupportingDocument[${index}]/WriteOff the element AmountAmount has been added with value ${amount}
    In the section SupportingDocument[${index}]/WriteOff the element AmountAmount has attribute currencyID with value ${currency}
```

**Expected behavior**
In keywords section I would expect variables to be highlighted as such

**Screenshots**
![image](https://user-images.githubusercontent.com/33200263/155600267-be4bbce8-41ac-44d2-abc3-6309d7476d54.png)

**Versions:**
 - OS: Windows 10
 - Robot Framework Version: [4]
 - Robot Framework Language Server Version: [0.41.0]
 - Client Version: [e.g.: VSCode 1.64.2]

PR #589, opened 2022-02-27
Implemented highlighting for embedded variables #586

weltings:
Embedded variables were not highlighted in BDD type statements like:

Financial transaction ${type} should be completed with amount ${amount}

This is now implemented, see also #586


fabioz:
It doesn't seem like this pull request is correct... 

It's including the code for what was already merged. It's the same branch you pushed previously (and it has the same code that's already in `master`).

Note that the proper approach is checking out your `master` branch (which should always try to keep in sync with the `master` branch from https://github.com/robocorp/robotframework-lsp because usually your code will be squashed to a single commit when merging, so, you can't keep using the same branch) then update it (this can be done with `git pull https://github.com/robocorp/robotframework-lsp.git master`), then create a new branch, do your changes there and provide a pull request for that new branch.

As there's no new code in this branch I'm going to close it (waiting for the proper pull request...).

fabioz:
p.s.: as you probably do have those commits locally, do the steps above to get a clean branch based on `master` and then use `git cherry-pick` to get the related commits in your new branch.

weltings:
Thanks for the tips

PR #590, opened 2022-02-28
Implemented support for embedded variables in BDD

weltings:
Embedded variables like the following:

User ${name} has received an email with subject ${subject}
Transaction ${type} has been completed with ${amount} and reference ${reference}

Will now be highlighted

### DIFF
--- a/robotframework-ls/src/robotframework_ls/impl/semantic_tokens.py
+++ b/robotframework-ls/src/robotframework_ls/impl/semantic_tokens.py
@@ -51,44 +51,55 @@ del value
 
 
 class _DummyToken(object):
-    __slots__ = ["type", "value", "lineno", "col_offset", "end_col_offset"]
+    __slots__ = ["type", "value", "lineno", "col_offset", "end_col_offset", "index"]
 
-    def __init__(self, initial_token=None):
+    def __init__(self, initial_token=None, index=None):
         if initial_token:
             self.type = initial_token.type
             self.value = initial_token.value
             self.lineno = initial_token.lineno
             self.col_offset = initial_token.col_offset
             self.end_col_offset = initial_token.end_col_offset
+            self.index = index
 
-    def extract_token(self, type, position):
+    def extract_token(self, type, position, trailing_whitespaces=1):
         extracted_token = _DummyToken()
         extracted_token.type = type
         extracted_token.value = self.value[:position]
         extracted_token.lineno = self.lineno
         extracted_token.col_offset = self.col_offset
         extracted_token.end_col_offset = self.col_offset + len(extracted_token.value)
-        self.__remove(extracted_token)
+        self.__remove(extracted_token, trailing_whitespaces)
         return extracted_token
 
-    def __remove(self, extracted_token):
+    def __remove(self, extracted_token, trailing_whitespaces):
         extracted_token_length = (
             extracted_token.end_col_offset - extracted_token.col_offset
         )
-        self.value = self.value[extracted_token_length + 1 :]
-        self.col_offset = extracted_token.end_col_offset + 1
+        self.value = self.value[extracted_token_length + trailing_whitespaces :]
+        self.col_offset = extracted_token.end_col_offset + trailing_whitespaces
+
+    def __eq__(self, other):
+        if (
+            self.col_offset == other.col_offset
+            and self.end_col_offset == other.end_col_offset
+        ):
+            return True
+        else:
+            return False
 
 
 def extract_gherkin_token_from_keyword(keyword_token):
-    gherkin_token = None
     import re
 
+    gherkin_token = None
     result = re.match(
         r"^(Given|When|Then|And|But)\s", keyword_token.value, flags=re.IGNORECASE
     )
     if result:
         gherkin_token_length = len(result.group(1))
         gherkin_token = keyword_token.extract_token("control", gherkin_token_length)
+        gherkin_token.index = TOKEN_TYPE_TO_INDEX[gherkin_token.type]
     return gherkin_token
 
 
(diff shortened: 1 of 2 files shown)


robotframework-ls/src/robotframework_ls/impl/semantic_tokens.py
```diff
@@ -356,17 +443,18 @@ def semantic_tokens_full(context: ICompletionContext):
         ast = context.doc.get_ast()
     except:
         return []
+    return semantic_tokens_full_from_ast(ast, context.monitor)
```
fabioz, line 446: This change was recently done and was undone in this commit (so, the changes which add back `semantic_tokens_full_from_ast` should be reverted).

robotframework-ls/src/robotframework_ls/impl/semantic_tokens.py
```diff
@@ -97,9 +108,101 @@ def extract_library_token_from_keyword(keyword_token):
...
+        )
+        bdd_statement_token.index = RF_TOKEN_TYPE_TO_TOKEN_TYPE_INDEX[KEYWORD]
+    return bdd_statement_token
+
+
+def extract_embedded_variable(keyword_token):
+    import re
+
+    embedded_variable_token = None
+    result = re.match(r"([$&@]{[^}]*?})", keyword_token.value)
```
fabioz, line 138: This isn't entirely correct because variables may be nested.

i.e.: in the code below `${OUTPUT${DIR}}` and `${OUTPUT_DIR}` are the same thing...

```
*** Test Cases ***
My Test
    ${output}=    Set Variable    OUTPUT_
    ${dir}=    Set Variable    DIR
    Log To Console    ${OUTPUT${DIR}}
    Log To Console    ${OUTPUT_DIR}
```

You need to do something as:

```
    token = ast_utils.create_token(text)
    try:
        tokenized_vars = ast_utils.tokenize_variables(token)
        for v in tokenized_vars:
            if v.type == v.VARIABLE:
            ...
    except VariableError:  # May throw error if it's not valid
       ...
```

We can initially not do it recursively to get the sub-variables if you feel it's too much work to do the recursion and get sub-variables at this point, but we should try to structure keeping that in mind...

robotframework-ls/src/robotframework_ls/impl/semantic_tokens.py
```diff
@@ -97,9 +108,101 @@ def extract_library_token_from_keyword(keyword_token):
...
+            tokens_parsed_recursively = extract_tokens_from_bdd_statement(
+                token_being_parsed
+            )
+            list_with_subtokens = list_with_subtokens + tokens_parsed_recursively
+    else:
+        list_with_subtokens = [token_being_parsed]
+    return list_with_subtokens
+
+
+def keyword_contains_embedded_variable(keyword_token):
```
fabioz, line 196: This method should be replaced by: `robotframework_ls.impl.text_utilities.contains_variable_text` -- although maybe ideally we should just do the call to `ast_utils.tokenize_variables(token)` just checking if `{` is present (because the tokenization is slow and if there are no tokens it'll just return the initial token and there'll be no variables in it).

robotframework-ls/src/robotframework_ls/impl/semantic_tokens.py
```diff
@@ -179,16 +282,22 @@ def _tokenize_token(node, initial_token):
...
-            yield token_gherkin_prefix, TOKEN_TYPE_TO_INDEX[token_gherkin_prefix.type]
+            yield token_gherkin_prefix, token_gherkin_prefix.index
         token_library_prefix = extract_library_token_from_keyword(token_keyword)
         if token_library_prefix:
-            yield token_library_prefix, TOKEN_TYPE_TO_INDEX[token_library_prefix.type]
-        if token_gherkin_prefix or token_library_prefix:
-            yield token_keyword, RF_TOKEN_TYPE_TO_TOKEN_TYPE_INDEX[KEYWORD]
-            return
+            yield token_library_prefix, token_library_prefix.index
+        is_bdd_statement = keyword_contains_embedded_variable(token_keyword)
```
fabioz, line 294: There's no need to check if it contains, just using `ast_utils.tokenize_variables(token)` as explained in other comments should already extract things appropriately without the check.

robotframework-ls/src/robotframework_ls/impl/semantic_tokens.py
```diff
@@ -356,17 +443,18 @@ def semantic_tokens_full(context: ICompletionContext):
...
+def semantic_tokens_full_from_ast(ast, monitor: Optional[IMonitor]):
+    from robotframework_ls.impl import ast_utils
 
     ret: List[int] = []
     append = ret.append
 
     last_line = 0
     last_column = 0
-    for _stack, node in ast_utils.iter_all_nodes_recursive(ast):
+    for _stack, node in ast_utils._iter_nodes(ast, recursive=True):
```
fabioz, line 457: This should also be undone (it should use the public API which was recently added).

robotframework-ls/tests/robotframework_ls_tests/test_semantic_highlighting_bdd.py
```diff
@@ -0,0 +1,199 @@
+from robotframework_ls.impl.completion_context import CompletionContext
```
fabioz, line 1: All the imports should be made locally and not globally in this module.

fabioz (changes requested):
I think most changes are related to the fact that regexps can't really be used to parse variables because they don't take the case of nested variables into account (the robot api should be used instead and that case needs to be accounted for).

robotframework-ls/src/robotframework_ls/impl/semantic_tokens.py
weltings: Fixed

robotframework-ls/src/robotframework_ls/impl/semantic_tokens.py
weltings: Fixed

robotframework-ls/tests/robotframework_ls_tests/test_semantic_highlighting_bdd.py
weltings: Fixed

fabioz:
@weltings do you still plan to tackle this? Otherwise I'll probably take a look at it...

fabioz:
Closing as this was implemented in: https://github.com/robocorp/robotframework-lsp/commit/0eeeca5ee36d6c6dc5247c48e3ff817ae381cd9a

PR #593, opened 2022-03-02
Add windows locator support for inspector

mmokko:
Update the robocorp-inspector to latest version (0.6.0).
New locator menu item: Add Windows locator.

@fabioz do we have any option to disable a menu item / command in certain platforms? This feature in inspector is Windows only. I added an error message for the user, but of course it would be better if the menu item is disabled.


fabioz:
Just merged. Related to hiding the button, that's possible... You should be able to use when `when` clause for that.

According to https://code.visualstudio.com/api/references/when-clause-contexts it seems there's a `isWindows` which can be used.

PR #750, opened 2022-09-06
intellij: update Gradle and intellij

KotlinIsland:
Hi there.

Here is a small PR updating the build script of the intellij plugin (just bumping `org.jetbrains.intellij` and the Gradle version).

@fabioz What is your opinion of converting the Gradle build script to `kts`? I'd be happy to do this, I've converted many projects from groovy to kts.

Also, have you seen [the intellij plugin template](https://github.com/JetBrains/intellij-platform-plugin-template)? It's got some great features in it, and I would be happy to implement some of those features into this project (I have also successfully done this to several projects)

Lastly, what's your opinion of using Kotlin instead of Java? I'd also be very happy to convert the source code of the plugin to Kotlin.


fabioz:
> Here is a small PR updating the build script of the intellij plugin (just bumping `org.jetbrains.intellij` and the Gradle version).

Thank you for the PR! Just merged.

 
> @fabioz What is your opinion of converting the Gradle build script to `kts`? I'd be happy to do this, I've converted many projects from groovy to kts.

If you'd like to take a look at this and you think the final result would be better, feel free to do it (personally I don't really care much for either way, both work fine for me).

> Also, have you seen [the intellij plugin template](https://github.com/JetBrains/intellij-platform-plugin-template)? It's got some great features in it, and I would be happy to implement some of those features into this project (I have also successfully done this to several projects)

Likewise, if you'd like to add other features from the default template, feel free to do it and provide a PR ;)

> Lastly, what's your opinion of using Kotlin instead of Java? I'd also be very happy to convert the source code of the plugin to Kotlin.

I must say I have much more experience in Java than in Kotlin and so far I haven't really bothered myself to deep dive more into Kotlin (although I played a bit with it and can understand it just fine in general), so, while I know it has some bells and whistles over Java I'm not sure I'd like to convert just for the sake of the conversion for the code itself -- personally I find java quite readable even thought it leans more on the verbose side... Although I go for Python when possible ;).

Still, my time for dealing with plugin for Intellij has been quite limited recently. I think it shows a bit that support for VSCode in the language server is better for VSCode than for the Intellij client as I haven't been able to focus the amount needed to polish it more -- I do occasional fixes and improvements on the Intellij client but I don't currently have the time to polish it more in the short/mid term as there's more focus on the language server itself at this point... thankfully many improvements on the backend benefit any clients, including intellij ;) .

Now, if you'd like to take a role that's more active in the project and contribute other improvements that'd definitely be something nice, so, if coding in Kotlin would be a pre-requisite for you to contribute those, what could be done is that your new contributions could be done in Kotlin (so, the project could mix java/kotlin there), but I wouldn't like to do a full conversion at this time (this could be done in the future if you'd like to become a champion for the Intelilj client though).

KotlinIsland:
Thanks for the reply!, I'd be happy to put together some PRs where I can, and we will see where things lead I guess. 😄

PR #751, opened 2022-09-07
DEVVS-66 | New Model Builder in VSCode 4 Flow Explorer

custompointofview:
Adding additional blocks to the model - if (improvements), while, for, try. 
Adding unittests for all.
Improving Flow Explorer

## TASK
https://linear.app/robocorp/issue/DEVVS-66/new-model-builder-in-vscode-4-flow-explorer


robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -49,9 +49,7 @@ def build_flow_explorer_model(completion_contexts: List[ICompletionContext]) ->
                     "body": test_body,
                 }
                 tasks.append(test)
-                for node_info in ast_utils.iter_all_nodes(
-                    test_node_info.node, recursive=False
-                ):
+                for node_info in ast_utils.iter_all_nodes(test_node_info.node, recursive=False):
```
fabioz, line 52: Do you have different configurations for black (or maybe a different black version)? 
-- it must be the one specified in `robocorp-code/dev_requirements.txt`

The CI is complaining about the formatting changes...

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -20,7 +20,7 @@ def build_flow_explorer_model(completion_contexts: List[ICompletionContext]) ->
         ast = completion_context.get_ast()
 
         # Uncomment to print ast.
-        # ast_utils.print_ast(ast)
+        ast_utils.print_ast(ast)
```
fabioz, line 23: Note to comment again before finishing...

fabioz (approved):
Seems great as is... 

We could do an intermediary commit with what's here (just need to comment back a few things and make black behave as intended).

fabioz:
Changes seem fine.

The failing test seems to be just because the generated json changed a bit (so, it's mostly a matter of updating that file).

We should add a test which checks for keywords in a different file (the code should handle it already, but it'd be nice to have a test and check that the expected result is what's expected by the flow explorer).

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -89,64 +142,92 @@ def _build_hierarchy(
...
+
+                        # If it was found in a library we don't recurse anymore.
+                        keyword_ast = keyword_found.keyword_ast
+                        if keyword_ast is None:
+                            continue
+                        definition_completion_context = keyword_found.completion_context
+                        if definition_completion_context is None:
+                            continue
+                        # If found in recursion stack we don't recurse anymore.
+                        if recursion_stack and recursion_stack.check(keyword["name"]):
```
fabioz, line 190: I don't think you need to check `if recursion_stack` part here (it can't be None as far as I can see so that'd always evaluate to `True` right?).

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -89,64 +142,92 @@ def _build_hierarchy(
...
+                            continue
+
+                        suite_name = _compute_suite_name(definition_completion_context)
+                        # Ok, it isn't a library keyword (as we have its AST). Keep recursing.
+                        for node_info in ast_utils.iter_all_nodes(
+                            keyword_ast, recursive=False
+                        ):
+                            if recursion_stack:
```
fabioz, line 198: I don't think you need to check `if recursion_stack` part here (it can't be None as far as I can see so that'd always evaluate to `True` right?).

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -89,64 +142,92 @@ def _build_hierarchy(
...
+                                recursion_stack.push(keyword["name"])
```
fabioz, line 199: Usually I do a `contextmanager` when there's something that needs a `push` / `pop`.

```
class _RecursionStack:

    ....

    @contextmanager
    def scoped(self, keyword_name)
        self.push(keyword_name)
        yield
        self.pop()
```

So it can be used as:

```
with recursion_stack.keyword_scope(keyword_name):
    ...
```

-- `push` and `pop` could probably be inlined in this case.

That's falls on the stylistic camp though, so, feel free to skip the suggestion.

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -4,6 +4,22 @@
 from robocorp_ls_core.basic import isinstance_name
 
 
+class _RecursionStack:
+    stack: List[str] = []
```
fabioz, line 8: You want to set the `stack` under the constructor (i.e.: `in def __init__(self)`) -- as it is any instance will use the same stack because it's declared in the class scope.

Also, it should probably be at least protected (i.e.: `_stack`).

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -37,29 +54,63 @@ def build_flow_explorer_model(completion_contexts: List[ICompletionContext]) ->
...
+                user_keyword_info = {
+                    "type": "user-keyword",
+                    "name": user_keyword_name,
+                    "doc": "",
+                    "body": user_keyword_body,
+                }
+
+                # Keywords var will be populated when building hierarchy if importing statements
+                # Checking to see if it already exists before appending
+                for keyw in keywords:
```
fabioz, line 94: We can leave as is for now, but I think we could have a better approach here as looping to check for contains is not ideal (I can see it becoming something slow if we reach the 500+ keywords use case).

-- keywords could be some custom structure which has a that keywords list and an additional `set` with the name of the keywords added so far where we could just do a `user_keyword_name in container`.

p.s.: just noting, we should do this later not in this commit.

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -74,6 +125,8 @@ def _build_hierarchy(
     suite_name: str,
     parent_body: List[Any],
     memo: dict,
+    recursion_stack: _RecursionStack,
+    user_keywords: List[Any] = None,
```
fabioz, line 129: This type isn't correct. If `user_keywords` may be `None`, it should be: `Optional[List[Any]]`.

Still, I don't follow it well, why is it `None` at all?

In the code the keyword is added to that list when it's found from a keyword call, but why doesn't it add it when it's a keyword inside a `for` or an `if`?

I'm having a bit of a hard time understanding the logic on when should a keyword be added there and when it shouldn't be added there. Can you elaborate a bit?

fabioz (changes requested):

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -1,9 +1,53 @@
...
 from robocorp_ls_core.basic import isinstance_name
 
 
+class _KeywordRecursionStack:
+    _stack: Dict
+
+    def __init__(self) -> None:
+        self._stack = {}
+
+    def has(self, keyword_name: str) -> bool:
```
fabioz, line 14: The pythonic thing here would probably be changing `has` to `__contains__` and then doing `if keyword_name in recursion_stack`.

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -1,9 +1,53 @@
...
+        return keyword_name in self._stack
+
+    def push(self, keyword_name: str) -> None:
+        self._stack.update({keyword_name: None})
```
fabioz, line 18: The idiom for this with a dict would be:

`self._stack[keyword_name] = None`

Still if you change the stack for a `set` it'd be `self._stack.add(keyword_name)`.

Additional notes:
- since `pop` would remove the entry, I think you should assert that you don't double-push.
- since this `push` should only be used inside the class it should be protected (or it could be inlined).

i.e.:

```
def _push(self, keyword_name: str):
    assert keyword_name not in self._stack
    self._stack.add(keyword_name)
```

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -1,9 +1,53 @@
...
 from typing import List, Any, Dict, Tuple
 import os
 from robocorp_ls_core.basic import isinstance_name
 
 
+class _KeywordRecursionStack:
+    _stack: Dict
+
+    def __init__(self) -> None:
+        self._stack = {}
```
fabioz, line 12: Given that you don't use the `dict` value for anything the `_stack` should be a `set`.

The typing for it would be `Set[str]`.

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -1,9 +1,53 @@
...
+
+    def has(self, keyword_name: str) -> bool:
+        return keyword_name in self._stack
+
+    def push(self, keyword_name: str) -> None:
+        self._stack.update({keyword_name: None})
+
+    def pop(self, keyword_name: str) -> None:
```
fabioz, line 20: Like `push`, `pop` should be protected i.e.: `_pop` (or it could just be inlined).

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -1,9 +1,53 @@
...
+    @contextmanager
+    def scoped(self, keyword_name):
+        self.push(keyword_name)
+        yield
+        self.pop(keyword_name)
+
+
+class _UserKeywordCollector:
+    _user_keywords: List[Any]
+    _name_collection: Dict
```
fabioz, line 32: This should be `Dict[str]`.

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -1,9 +1,53 @@
...
+
+    def __init__(self) -> None:
+        self._user_keywords = []
+        self._name_collection = {}
+
+    @property
+    def keywords(self) -> List[Any]:
+        return self._user_keywords
+
+    def has(self, keyword_name) -> bool:
```
fabioz, line 42: Typing for `keyword_name: str`.

Also, the pythonic thing here would be doing a `__contains__` and doing a `keyword_name in container`.

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -1,9 +1,53 @@
...
+        self.pop(keyword_name)
+
+
+class _UserKeywordCollector:
+    _user_keywords: List[Any]
+    _name_collection: Dict
+
+    def __init__(self) -> None:
+        self._user_keywords = []
+        self._name_collection = {}
```
fabioz, line 36: `_name_collection` should be a `Set[str]` instead of a dict.

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -1,9 +1,53 @@
...
+    def keywords(self) -> List[Any]:
+        return self._user_keywords
+
+    def has(self, keyword_name) -> bool:
+        return keyword_name in self._name_collection
+
+    def append(self, keyword: Any) -> None:
+        if "name" in keyword:
+            self._user_keywords.append(keyword)
+            self._name_collection.update({keyword["name"]: None})
```
fabioz, line 48: After changing `_name_collection` for a `set` use `self._name_collection.add(keyword['name`]).

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -1,9 +1,53 @@
...
+        self._name_collection = {}
+
+    @property
+    def keywords(self) -> List[Any]:
+        return self._user_keywords
+
+    def has(self, keyword_name) -> bool:
+        return keyword_name in self._name_collection
+
+    def append(self, keyword: Any) -> None:
```
fabioz, line 45: I think the typing could be a bit more specific here (`keyword: dict`).

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -1,9 +1,53 @@
...
+
+    @contextmanager
+    def scoped(self, keyword_name):
+        self.push(keyword_name)
+        yield
+        self.pop(keyword_name)
+
+
+class _UserKeywordCollector:
+    _user_keywords: List[Any]
```
fabioz, line 31: Typing could probably be `List[dict]`.

Sidenote (no need to do):

The actual keyword dict itself could be typed with a `TypedDict` so it could be `List[UserKeywordTypedDict]` (but then all the references to it would need to be typed too, not sure if it's worth the effort in this case).

See: https://peps.python.org/pep-0589/

robotframework-ls/src/robotframework_ls/impl/flow_explorer_model_builder.py
```diff
@@ -37,29 +83,61 @@ def build_flow_explorer_model(completion_contexts: List[ICompletionContext]) ->
...
+                user_keyword_info = {
+                    "type": "user-keyword",
+                    "name": user_keyword_name,
+                    "doc": "",
+                    "body": user_keyword_body,
+                }
+
+                # Keywords var will be populated when building hierarchy if importing statements
+                # Checking to see if it already exists before appending
+                for keyw in keywords:
```
fabioz, line 122: I think you wanted to do the following here: 

```
if keyw['name'] in user_keywords_collector:
    break
else:
    user_keywords_collector.append(user_keyword_info)
    ...
```

(if you change the `has` for `__contains__`).

fabioz (commented):
Some notes on making things a bit more pythonic.

Also, I think we may have a problem with keyword names if they're used with different casing (so, this should at least have a test).

i.e.:

What happens if the user does:

```
*** Keywords ***
my keyword
    no operation

*** Test Cases ***
My test
    my keyword
    mykeyword
    My Keyword
```

Internally we should make sure we don't generate additional entries (as all the related calls point to the same definition).

There are some utilities for this...
`robotframework_ls.impl.text_utilities.normalize_robot_name` can be used to get the name as is seen by robotframework internally to make matches.

The final name should usually be the one that the user has used as the real keyword name, but for libraries (where we generate the name) there's a utility which will create the proper name.

See: `create_convert_keyword_format_func` (it receives the `completion_context.config` and returns a callable which can be called with the keyword name from the library and generates the one that should be shown to the user).

fabioz (approved):
Nice work ;)

PR #754, opened 2022-09-13
Add Inspectors recorder to palette

xylix:
Notes:
- The internal command `robocorp.newRobocorpInspectorRecord` should probably be changed to something more clear
- The `inspector record` command is very user interaction dependant, so might be difficult to write unit tests for it
- When running the code currently the palette command is failing with `Command 'Robocorp: Record series of browser actions' resulted in an error (An unknown error occurred. Please consult the log for more details.)`. I'm having trouble even finding the relevant logs. Help with this or guesses about the problem are very welcome.


fabioz:
Usually when VSCode refers to logs it's something that may be viewed in the `OUTPUT` tab.

So, it'd be `OUTPUT > Robocorp Code` (but depending on what happened it could be another tab and if you're running with the debugger it could end up being shown in the VSCode `Debug Console`).

If you want we can do a zoom about it.

xylix:
@fabioz When I'm running it from one VSCode window the debug console and output seem to both be empty, both in the "Extension Development Host" window and the other window.
Screenshot 2022-09-14 at 14 34 16

fabioz:
@xylix do you have plans to work on this on the short term?

i.e.: if there aren't plans to work on it on the short term I was thinking about closing and reopening when there's new code to review.

xylix:
Sorry, I was stuck for some time and didn't manage to make meaningful progress. @custompointofview might make some progress on this soon, so keeping this open for now.

custompointofview:
@fabioz This is ready to go! 😊

robocorp-code/vscode-client/src/inspector.ts
```diff
@@ -9,17 +9,34 @@ import { OUTPUT_CHANNEL } from "./channel";
...
+export enum InspectorType {
+    Browser = "browser",
+    Windows = "windows",
+    Image = "image",
+    WebRecorder = "web-recorder",
+}
+
+export type InspectorTypes = `${InspectorType}`;
+
+export const DEFAULT_INSPECTOR_VALUE = {
```
fabioz, line 21: I don't really see a reason for the `DEFAULT_INSPECTOR_VALUE` (since it'll always be `false` for everything)

robocorp-code/vscode-client/src/viewsLocators.ts
```diff
@@ -84,6 +85,8 @@ export class LocatorsTreeDataProvider
             iconPath = "error";
         } else if (type === "windows") {
             iconPath = "multiple-windows";
+        } else if (type === "web-recorder") {
```
fabioz, line 88: Since the `InspectorType` enum is now available, I guess that this should be `InspectorType.WebRecorder` (and the other related `ifs` should be updated too...).

robocorp-code/vscode-client/src/inspector.ts
```diff
@@ -9,17 +9,34 @@ import { OUTPUT_CHANNEL } from "./channel";
...
+    browser: false,
+    image: false,
+    windows: false,
+    "web-recorder": false,
+};
+
+let _openingInspector: { [K in InspectorTypes]: boolean } = DEFAULT_INSPECTOR_VALUE;
```
fabioz, line 28: I'm not sure about this change. It was this way to prevent users from opening multiple inspectors at the same time -- I'm not sure about the current version, but I remember that opening say a browser recorder and an image recorder at the same time, if changing one the other wouldn't really know about such changes and I believe the user could even loose one thing recorded in one of those (given that they all use the same `locators.json`).

So, unless we actually can guarantee that this isn't the case, I'm not sure we should allow for that per-inspector -- although if this is now a valid scenario that was tested for, then that'd be ok ;)

robocorp-code/bin/create_env/conda.yaml
```diff
@@ -1,9 +1,9 @@
 channels:
-  - conda-forge
-dependencies:
-  - python=3.8.10
-  - pip=20.1
-  - pip:
-    - robocorp-inspector==0.6.5
-    - robotframework==4.1
+    - conda-forge
```
fabioz, line 2: We need to update the base environment with the conda.yaml changes now (as commented on slack, we should add some comments to the `conda.yaml` files so that we don't forget about that...).

fabioz (changes requested):

robocorp-code/vscode-client/src/inspector.ts
custompointofview: The Inspector treats the `locators.json` as a database now. 
It keeps a lock on it while writing & reading. 
The Picker app doesn't need a more forceful approach when dealing with the file. 
The Web Recorder has a polling system that keeps the database up to date & the save ability doesn't overlap with the Picker.

The user has the ability (with these changes) to open the Picker & the Web Recorder at the same time - but just one instance of each.  It would be close to impossible to create a race condition when file handling the `locators.json`.

From all the testing, it doesn't seem to be a problem while the two instances are opened. 
One improvement could be to just open the apps in the same process. 

This solution superficially solved a problem with hanging processes. While closing the Inspector the process hanged in the background, and as such you couldn't open another Inspector app because the process was considered in a running state. Fixed the issue afterwards, but the solution helped.

If still have concerns over this, I can revert some of the changes & keep a single process for the entire Inspectors spectrum of apps.

fabioz: Seems good to me then ;)

PR #757, opened 2022-09-15
DEVVS-67 | Robocorp Converter Integration

custompointofview:
Integrating converter to the VSCode extension - adding command to convert active file to Robocorp project, adding ability to download the converter bundle, run the converter, gather resulting files.

### DIFF
--- a/robocorp-code/codegen/commands.py
+++ b/robocorp-code/codegen/commands.py
@@ -573,8 +573,8 @@ COMMANDS = [
         hide_from_command_palette=False,
     ),
     Command(
-        "robocorp.convertProject.internal",
-        "Convert file to Robocorp Project Internals",
+        "robocorp.saveConvertedProject.internal",
+        "Save converted Robocorp Project",
         add_to_package_json=True,
         server_handled=True,
         hide_from_command_palette=True,
--- a/robocorp-code/codegen/commands.py
+++ b/robocorp-code/codegen/commands.py
@@ -565,6 +565,20 @@ COMMANDS = [
         hide_from_command_palette=True,
         icon="$(type-hierarchy-sub)",
     ),
+    Command(
+        "robocorp.convertProject",
+        "Convert file to Robocorp Project",
+        add_to_package_json=True,
+        server_handled=False,
+        hide_from_command_palette=False,
+    ),
+    Command(
+        "robocorp.convertProject.internal",
+        "Convert file to Robocorp Project Internals",
+        add_to_package_json=True,
+        server_handled=True,
+        hide_from_command_palette=True,
+    ),
 ]
 
 
--- a/robocorp-code/codegen/commands.py
+++ b/robocorp-code/codegen/commands.py
@@ -567,7 +567,7 @@ COMMANDS = [
     ),
     Command(
         "robocorp.convertProject",
-        "Convert file to Robocorp Project",
+        "Convert file in active editor to Robocorp Project",
         add_to_package_json=True,
         server_handled=False,
         hide_from_command_palette=False,
--- a/robocorp-code/vscode-client/src/extension.ts
+++ b/robocorp-code/vscode-client/src/extension.ts
@@ -395,7 +395,7 @@ async function convertProject() {
             return;
         }
         const populate_action: ActionResult<string> = await commands.executeCommand(
-            "robocorp.convertProject.internal",
+            "robocorp.saveConvertedProject.internal",
             {
                 "destinationFolderURI": destination.toString(),
                 "conversionResult": conversionResult,
--- a/robocorp-code/vscode-client/src/extension.ts
+++ b/robocorp-code/vscode-client/src/extension.ts
@@ -136,13 +136,14 @@ import {
 import { installPythonInterpreterCheck } from "./pythonExtIntegration";
 import { refreshCloudTreeView } from "./viewsRobocorp";
 import { connectVault, disconnectVault } from "./vault";
-import { ensureConvertBundle, getLanguageServerPythonInfoUncached } from "./extensionCreateEnv";
+import { getLanguageServerPythonInfoUncached } from "./extensionCreateEnv";
 import { registerDebugger } from "./debugger";
 import { clearRCCEnvironments, clearRobocorpCodeCaches, computeEnvsToCollect } from "./clear";
 import { Mutex } from "./mutex";
 import { mergeEnviron } from "./subprocess";
 import { feedback } from "./rcc";
 import { showSubmitIssueUI } from "./submitIssue";
+import { ensureConvertBundle } from "./convertion";
 
 interface InterpreterInfo {
     pythonExe: string;
@@ -340,13 +341,18 @@ async function convertProject() {
             Could not convert project.
             Please check the output logs for more details.
             `;
-
     const DEFAULT_ERROR_STATUS = "Error while converting project.";
 
+    const activeTextEditor = window.activeTextEditor;
+    if (!activeTextEditor) {
+        window.showErrorMessage("Could not detect an active editor");
+        return;
+    }
+
     try {
-        const converterLocation = await ensureConvertBundle(true);
+        const converterLocation = await ensureConvertBundle();
         if (!converterLocation) {
-            throw new Error("Converter bundle was not found. Possible download issues.");
+            throw new Error("Converter bundle was not found. Possible download issues. Please try again.");
         }
         const converterBundle = require(converterLocation);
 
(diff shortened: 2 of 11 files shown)


robocorp-code/vscode-client/src/extension.ts
```diff
@@ -324,6 +326,102 @@ async function cloudLogoutAndRefresh() {
...
+            "UiPath": "uipath",
+            "Automation Anywhere 360": "a360",
+        };
+        const items = Object.keys(vendorMap);
+        let selectedItem = await window.showQuickPick(items, {
+            "placeHolder": "Please select vendor of current project",
+            "canPickMany": false,
+            "ignoreFocusOut": false,
+        });
+        window.setStatusBarMessage(`Selected project type: ${selectedItem}`);
```
fabioz, line 373: I think you shouldn't set `window.setStatusBarMessage` at all in this method (if the user just activated the action or pressed esc, etc, I don't think we should notify anything to him in the status bar).
-- the calls to information/error when done should already suffice.

You can use the `OUTPUT_CHANNEL` if you want to log something though (it will then be shown in `OUTPUT > Robocorp Code`), but I'm not sure we even need to log these (unless something went bad).

robocorp-code/vscode-client/src/extension.ts
```diff
@@ -324,6 +326,102 @@ async function cloudLogoutAndRefresh() {
...
+
+        // let the user decide where the conversion result will be saved
+        const destination = await window.showOpenDialog({
+            title: "Select conversion destination",
+            openLabel: "Select Output Folder",
+            canSelectFiles: false,
+            canSelectFolders: true,
+            canSelectMany: false,
+        });
+        if (!destination || destination.length == 0) {
```
fabioz, line 387: We should also check if we're about to override anything in the selected folder and ask him if overriding is ok (right now any file with `tasks%d.robot` could be overridden -- maybe we should change the output files to `converted_task%d.robot` to have less possible conflicts with user code).

I was thinking that maybe we could not even have this selection and just save to `${workspaceFolder}/converted` (we'd have to ask for the workspace folder if there are multiple though -- i.e.: `window.showWorkspaceFolderPick`), but always pick a new folder name automatically there (so, there would never be conflicts and the user wouldn't need to do a selection in the common case).

robocorp-code/vscode-client/src/extension.ts
```diff
@@ -324,6 +326,102 @@ async function cloudLogoutAndRefresh() {
...
+        if (!converterBundle.isSuccessful(conversionResult)) {
+            logError("Error while converting project", new Error(conversionResult.error), "EXT_CONVERT_PROJECT");
+            window.setStatusBarMessage(DEFAULT_ERROR_STATUS);
+            window.showErrorMessage(DEFAULT_ERROR_MSG);
+            OUTPUT_CHANNEL.show();
+            return;
+        }
+        const populate_action: { result: string; success: boolean; message: string | null } | null =
+            await commands.executeCommand("robocorp.convertProject.internal", {
+                "destinationFolderURI": destination[0].path,
```
fabioz, line 408: The var names says it's an `uri`, yet you're passing a `path`... this doesn't seem right.

In general you either pass the uri directly (just do a `toString()` with it) and then in the python side you need to do `robocorp_ls_core.uris.to_fs_path` to access the filesystem path in the language server or you pass a path -- i.e.: `destinationFolderPath` and pass the `uri.fsPath`.

Passing the `uri.toString()` is preferred though as most of the language server APIs deal with `uris` and not `paths` (it's converted only when needed).

robocorp-code/vscode-client/src/extension.ts
```diff
@@ -324,6 +326,102 @@ async function cloudLogoutAndRefresh() {
...
+            canSelectMany: false,
+        });
+        if (!destination || destination.length == 0) {
+            window.setStatusBarMessage("Project conversion was canceled.");
+            return;
+        }
+
+        window.setStatusBarMessage("Project conversion is starting...");
+        const activeTextEditor = window.activeTextEditor;
+        const conversionResult = await converterBundle.convert(
```
fabioz, line 394: It'd be nice to have some interface for the `conversionResult` to know what is actually there...

robocorp-code/vscode-client/src/extension.ts
```diff
@@ -324,6 +326,102 @@ async function cloudLogoutAndRefresh() {
...
+        );
+        window.setStatusBarMessage("Done converting project.");
+        if (!converterBundle.isSuccessful(conversionResult)) {
+            logError("Error while converting project", new Error(conversionResult.error), "EXT_CONVERT_PROJECT");
+            window.setStatusBarMessage(DEFAULT_ERROR_STATUS);
+            window.showErrorMessage(DEFAULT_ERROR_MSG);
+            OUTPUT_CHANNEL.show();
+            return;
+        }
+        const populate_action: { result: string; success: boolean; message: string | null } | null =
```
fabioz, line 406: We have an `ActionResult` interface for this (in `protocols.ts`).

fabioz (changes requested):

robocorp-code/vscode-client/src/extension.ts
```diff
@@ -324,6 +335,85 @@ async function cloudLogoutAndRefresh() {
...
+            // Operation cancelled.
+            return;
+        }
+        const destination = Uri.joinPath(ws.uri, "converted");
+        const activeTextEditor = window.activeTextEditor;
+        const conversionResult: ConversionResult = await converterBundle.convert(
+            vendorMap[selectedVendor],
+            activeTextEditor.document.getText()
+        );
+        if (!converterBundle.isSuccessful(conversionResult) && conversionResult.type === "Failure") {
```
fabioz, line 391: Should this be an `||` instead of `&&`?

robocorp-code/vscode-client/src/extension.ts
```diff
@@ -324,6 +335,85 @@ async function cloudLogoutAndRefresh() {
...
+            logError("Error while converting project", new Error(conversionResult.error), "EXT_CONVERT_PROJECT");
+            window.showErrorMessage(DEFAULT_ERROR_MSG);
+            OUTPUT_CHANNEL.show();
+            return;
+        }
+        const populate_action: ActionResult<string> = await commands.executeCommand(
+            "robocorp.convertProject.internal",
```
fabioz, line 398: I think that this action could have a different name (because the project is already converted at this point, it's just saving it now, so `robocorp.convertProject.internal` may be a bit misleading).

How about `robocorp.saveConvertedProject.internal`?

robocorp-code/vscode-client/src/extensionCreateEnv.ts
```diff
@@ -207,6 +278,13 @@ export async function createDefaultEnv(
         return;
     }
 
+    const convertBundleLocation = await ensureConvertBundle();
```
fabioz, line 281: We shouldn't do this here. We only want to download when the action is actually executed.

robocorp-code/vscode-client/src/extensionCreateEnv.ts
```diff
@@ -172,6 +182,67 @@ async function verifyLongPathSupportOnWindows(rccLocation: string): Promise<bool
     return true;
 }
 
+export async function ensureConvertBundle(justCheck: boolean = false) {
```
fabioz, line 185: Instead of a flag to `justCheck`, it should always do the check the first time it's called and update accordingly (but then, save in a global var if it had checked already and if everything is ok don't download anything anymore if the file is present locally and the previous check worked).

fabioz (changes requested):

robocorp-code/vscode-client/src/extension.ts
custompointofview: The operators are mostly equivalent, but because the bundle & its contents are hidden, Typescript will complain when dealing with values that are in either the ConversionSuccess | ConversionError.

robocorp-code/vscode-client/src/extensionCreateEnv.ts
custompointofview: This is where I figured we should download the bundle just once when the Extension is initialised & afterwards check the existence of the file.
Probably I got confused between these two comments: https://github.com/robocorp/robotframework-lsp/pull/757#discussion_r972272701

As the action has additional steps, it might be a bit tedious to wait for download to finish and move on. 
It seemed natural to do it in the env init as it is part of the environment functionality. 
But if you think I should move this, wouldn't be a problem on my end 😊

robocorp-code/codegen/commands.py
```diff
@@ -565,6 +565,20 @@ def __init__(
         hide_from_command_palette=True,
         icon="$(type-hierarchy-sub)",
     ),
+    Command(
+        "robocorp.convertProject",
+        "Convert file to Robocorp Project",
```
fabioz, line 570: As mentioned, we probably could have a different name to make the purpose clearer (let's do this over slack).

robocorp-code/vscode-client/src/extensionCreateEnv.ts
fabioz: The idea here is that most users will almost never use this feature, so, there's no point to everyone paying for that cost at activation time.

Also, we want to update the bundle when a new one comes out (but we should just do the update check once per session, so, just recheck for a new version after a VSCode restart).

robocorp-code/vscode-client/src/extensionCreateEnv.ts
fabioz: This is because we want to update the bundle when a new one comes out (but we should just do the update check once per session, so, just recheck for a new version after a VSCode restart).

robocorp-code/vscode-client/src/extension.ts
```diff
@@ -324,6 +335,85 @@ async function cloudLogoutAndRefresh() {
...
+
+        // let the user decide what type of project will be converted
+        const vendorMap = {
+            "Blue Prism": "blueprism",
+            "UiPath": "uipath",
+            "Automation Anywhere 360": "a360",
+        };
+        const items = Object.keys(vendorMap);
+        let selectedVendor = await window.showQuickPick(items, {
+            "placeHolder": "Please select the format of current project",
```
fabioz, line 361: I think the logic should first get the active editor (and validate it if it's valid as the user may not have an active editor) and then show in the title the name of the file that'll be converted (just so that the user know we're converting the active editor).

Also, not sure if it's possible, but if we know that the file extension (say, .xml), we could do some pre-validation and say that it's wrong from the get-go (i.e.: if the user opens a .robot file and asks for the conversion we already know it won't work...).

fabioz (changes requested):

robocorp-code/vscode-client/src/extension.ts
custompointofview: I've changed the logic.
But what should happen if there is no active editor? Should we open a file picker and open the file in the active editor and continue with the conversion? 🤔 

In regards to the file format auto-detection, we don't have an implemented solution yet. 
I did more of a superficial one once, but people were a bit against the idea as we don't know what extensions or formats of the exported 3rd parties projects. But I do think a task for this is in progress (or at least we should issue a reminder). The fail fast idea is still valid, but the format validation should be part of the converter so it becomes a common tool.

fabioz: Agreed, that validation should be a part of the converter.

robocorp-code/vscode-client/src/extension.ts
fabioz: I'm not sure I follow... if they're mostly the same, I guess you can just do: `!converterBundle.isSuccessful(conversionResult)` and not the `&& conversionResult.type === "Failure"` part, right?

PR #763, opened 2022-09-24
feat: add intellij line marker

jsmzr:
- Add intellij line marker

Quickly execute a case or file, set the console encoding to `utf-8`, set default log dir to `./log`.

robotframework-plugin-line-marker

### DIFF
--- a/robotframework-intellij/src/main/java/robocorp/dap/RobotRunAction.java
+++ b/robotframework-intellij/src/main/java/robocorp/dap/RobotRunAction.java
@@ -1,0 +1,80 @@
+package robocorp.dap;
+
+import com.intellij.execution.ExecutionException;
+import com.intellij.execution.ProgramRunnerUtil;
+import com.intellij.execution.RunManager;
+import com.intellij.execution.RunnerAndConfigurationSettings;
+import com.intellij.execution.configurations.RunConfiguration;
+import com.intellij.execution.executors.DefaultRunExecutor;
+import com.intellij.execution.runners.ExecutionEnvironment;
+import com.intellij.execution.runners.ExecutionEnvironmentBuilder;
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.project.ProjectManager;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class RobotRunAction extends AnAction {
+    private static final RobotConfigurationFactory FACTORY = new RobotConfigurationFactory(new RobotConfigurationType());
+    private final String name;
+    public RobotRunAction() {
+        this.name = null;
+    }
+    public RobotRunAction(String name) {
+        this.name = name;
+    }
+
+
+    @Override
+    public void actionPerformed(@NotNull AnActionEvent e) {
+        Project project = ProjectManager.getInstance().getOpenProjects()[0];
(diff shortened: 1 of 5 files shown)


fabioz (changes requested):
Hello, thank you very much for the PR.

I do have some comments though...

Although it handles `Test Cases`, it will fail with `Test case`  or `Task` / `Tasks`... and with translation which will be added in RF 5.1 it'll also fail with translations to other languages (because those names are translated in that case and it's hard to know the actual valid names -- unless the language server is asked for those, but that changes depending on a global setting or a per-file setting, so, it's not ideal to compute this in the client side since the language server does have all that knowledge already).

I think the best would be to actually ask the language server to compute that (since it'll consider all of those details).

The ideal API for this would be using the code lens API as this is actually computed using the code lens that has the name:

`robot.runTest` (with a name of `*` for running the test suite or the actual test).

The actual run is pretty much what you have implemented...

The way to go around this should be creating a `FeatureCodeLens` (like other `FeatureXXX` classes) which would query the information needed and store it (and we'd need to have some care not to query it again until the editor contents changed) and it should be able to map the regions from that to the psi element required in the client side (so, on the first request it'd do that query and afterwards it'd keep on reusing it unless the document itself changed).

This feature is also missing test cases (you can see existing tests for an inspiration). It can just create the provider directly and query it based on the psi for an editor you opened.

Some base test contents could be something as (add it to the `resources` folder and then you can open an editor in a test case with: `myFixture.configureByFile("caserunlinemarks.robot");`:

```
*** Test Cases ***
First test
    No Operation

Second test
    FOR    ${counter}    IN RANGE    START    END    opt.STEPS
        Log    ${counter}
    END

# Comment
# Comment
Third Test
    No Operation

*** Keywords ***
My Keyword
    No Operation

*** Test Case ***
Fourth Test
    No Operation

# Comment

Fifth Test
    No Operation
```

I know it changes quite a bit the implementation, so, please let me know if you think you'll be able to tackle it.

If you think that approach will take too much time (or if you're not up to doing it), I can do an initial commit with the current approach, but it should at least contain a test and support test case/test cases/task/tasks.

The match to know if it's a psi element for a test case/task header should probably be something as:

```
    private boolean isTestCaseElement(@NotNull PsiElement element) {
        if (!(element instanceof LSPPsiAstElement)) {
            return false;
        }
        LSPPsiAstElement lspPsiElement = ((LSPPsiAstElement) element);
        ASTNode node = lspPsiElement.getNode();
        IElementType elementType = node.getElementType();
        if (elementType != RobotElementType.HEADING) {
            return false;
        }

        String text = element.getText().strip();
        if (text.length() == 0) {
            return false;
        }
        text = text.replace('*', ' ').toLowerCase().trim();
        if (text.equals("test case") || text.equals("task") || text.equals("test cases") || text.equals("tasks"))  {
            return true;
        }
        return false;
    }
```

jsmzr:
Thank you very much for your suggestion, major modifications are difficult for me(I didn't know much about robotframework). I'll refer to your code to make modifications and add some tests.

fabioz:
This was merged in: https://github.com/robocorp/robotframework-lsp/commit/bc1fb8ebee23f2957df8885103af5b34b14972de

I ended up doing some additional changes after the merge because there were still some issues recognizing test cases in some situations (and in the end I added support for a right-click to debug too).

I guess the only thing really missing in the feature would be for users to be able to customize the template being used for the run so that he could have things such as environment variables or a custom command line -- but this can be done in the future ;)

Thank you very much for the PR (this will be available in version 1.2.0 -- I've just done the release but it can take some time before it appears as being available in the Intellij plugins store... it's already available in https://plugins.jetbrains.com/plugin/16086-robot-framework-language-server/versions/stable/230148 though).

PR #773, opened 2022-10-03
CONV-139 Download converter commons and use convertyml when calling converter

mstn:
* Download converter js AND commons in a single zip file
* Call converter js with (optional) common implementation (convert.yaml)

### DIFF
--- a/robocorp-code/package.json
+++ b/robocorp-code/package.json
@@ -955,22 +955,22 @@
         "prettier-fix": "npx prettier -w vscode-client/**/*.ts"
     },
     "devDependencies": {
-        "@types/adm-zip": "^0.5.0",
         "@types/mocha": "^2.2.32",
         "@types/node": "^13.0.00",
-        "@types/rimraf": "^3.0.2",
         "@types/vscode": "1.65.0",
+        "@types/adm-zip": "^0.5.0",
+        "@types/rimraf": "^3.0.2",
         "prettier": "2.4.1",
-        "typescript": "^4.5.4",
-        "vscode-test": "1.5.1"
+        "vscode-test": "1.5.1",
+        "typescript": "^4.5.4"
     },
     "dependencies": {
-        "adm-zip": "^0.5.9",
+        "vscode-languageclient": "^7.0.0",
+        "path-exists": "^4.0.0",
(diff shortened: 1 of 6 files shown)


robocorp-code/package.json
```diff
@@ -955,18 +955,22 @@
...
+        "typescript": "^4.5.4",
+        "vscode-test": "1.5.1"
     },
     "dependencies": {
-        "vscode-languageclient": "^7.0.0",
-        "path-exists": "^4.0.0",
+        "adm-zip": "^0.5.9",
         "http-proxy-agent": "^2.1.0",
         "https-proxy-agent": "^2.2.4",
+        "path-exists": "^4.0.0",
```
fabioz, line 971: The package.json can't be changed directly. Rather, edit `robocorp-code/codegen/codegen_package.py` to add/change dependencies (and then rerun that script in python).

fabioz (changes requested):
I think this isn't the ideal workflow. We should download only a single file.

Ideally this would be a single .js bundle with everything.

If having a single .js bundle with everything isn't possible due to some technical limitation and a .zip is indeed really needed, then it should be a single download for the .zip.

We can sync on on Slack (I guess I don't know enough about to structure to understand what'd be the limitations to having a single .js bundle).

Also, we must take care not to break existing clients (i.e.: in the released version we just use the single .js bundle -- if this isn't working right now we should probably revert so that it keeps working with an older version and provide a new url for this new download).

fabioz:
@mstn can you sign the contributor license agreement so that the bot is happy?

robocorp-code/vscode-client/src/rcc.ts
```diff
@@ -215,6 +215,7 @@ export async function download(
             let response: XHRResponse = await xhr({
                 "url": url,
                 "onProgress": onProgress,
+                timeout: 5000 /* 5 sec */,
```
fabioz, line 218: I think that this timeout isn't ideal... the same function is used to download RCC (9MB), so, this could timeout too easily in valid cases depending on the speed connection.

So, if we need a limit it should be much bigger by default (or we should provide this as an argument and customize it just for this use case).

fabioz (changes requested):

PR #819, opened 2022-11-26
optimize index order in large project

Zoupers:
In large project, a full index takes a while, but we can still run as usual.


Zoupers:
need to change some information

PR #829, opened 2022-12-05
Add default output dir in the config snippet

pasikokko:
Adding the output folder in the launch configuration snippets, to make sure output files are always generated in the same folder, despite how you start the robot.

### DIFF
--- a/robotframework-ls/codegen/codegen_package.py
+++ b/robotframework-ls/codegen/codegen_package.py
@@ -262,7 +262,10 @@ def get_json_contents():
                                 "target": '^"\\${file}"',
                                 "terminal": "integrated",
                                 "env": {},
-                                "args": [],
+                                "args": [
+                                    "--outputdir",
+                                    "output"
+                                ],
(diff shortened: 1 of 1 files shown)


fabioz:
If it should be the same dir it should probably use the `${workspaceFolder}` variable so that it doesn't rely on the `cwd`.

Also, please run the `codegen_package.py` so that the `package.json` file is regenerated with those updated values.

pasikokko:
Actually, the output dir should be inside the robot folder, but I haven't figured out how to configure that in generic way, so that it would work for many robots in the workspace

fabioz:
Humm, that's really trickier. The problem is that the `Robot Framework Language Server` doesn't really have much information on the actual structure of `Robocorp Code` in this case.

I'll have to think a bit more on how to solve this properly as this connection needs to be thought a bit better in this case. Can you create a ticket in: https://github.com/robocorp/robotframework-lsp/issues with the actual use case?

PR #877, opened 2023-02-01
Specify encoding when saving file on disk

mstn:
Converter returns now also info on file encoding when an encoding different from `utf-8` should be used.
* If converter output specifies an encoding, use that encoding
* Otherwise default to `utf-8`.

### DIFF
--- a/robocorp-code/vscode-client/src/conversion.ts
+++ b/robocorp-code/vscode-client/src/conversion.ts
@@ -360,9 +360,13 @@ export async function convertAndSaveResults(
 
                 const filesWritten: string[] = [];
 
-                async function handleOutputFile(file: string, content: string): Promise<void> {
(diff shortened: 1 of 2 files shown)


robocorp-code/vscode-client/src/conversion.ts
```diff
@@ -374,7 +378,13 @@ export async function convertAndSaveResults(
...
+                            tasks.push(
+                                handleOutputFile(
+                                    join(result.outputDir, f.filename),
+                                    f.content,
+                                    f.encoding as BufferEncoding
+                                )
+                            );
                         }
                     }
                     const images = okResult?.images;
```
fabioz, line 390: A question: should the `handleOutputFile` for the images have `binary` by default?

(do we even use those?)

PR #878, opened 2023-02-02
Create new converter command: schema

mstn:
Schema and analysis are two different things. Schema is a representation of the grammar of a RPA language that is used to improve the converter while analysis is a report intended for end users.

Schema is also slower in general. Having it isolated allows us to run it optionally.


fabioz (approved):
Seems good to me ;)

PR #881, opened 2023-02-06
UiPath: add command for generating schemas

mstn:
~~Note: don't merge, I have not pushed the new version of the converter yet.~~

### DIFF
--- a/robocorp-code/vscode-client/src/conversion.ts
+++ b/robocorp-code/vscode-client/src/conversion.ts
@@ -209,7 +209,7 @@ export async function convertAndSaveResults(
                     projects: projects,
                     onProgress: undefined,
                     outputRelativePath: join(nextBasename, basename(it)),
-                })
+                });
                 for (const it of opts.input) {
                     rpaConversionCommands.push({
                         vendor: Format.UIPATH,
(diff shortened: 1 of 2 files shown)


robocorp-code/vscode-client/src/conversion.ts
```diff
@@ -200,8 +200,16 @@ export async function convertAndSaveResults(
...
-            case RPATypes.uipath:
+            case RPATypes.uipath: {
+                const projects: Array<string> = opts.input;
                 nextBasename = await findNextBasenameIn(opts.outputFolder, "converted-uipath");
+                rpaConversionCommands.push({
+                    command: CommandType.Schema,
+                    vendor: Format.UIPATH,
+                    projects: projects,
+                    onProgress: undefined,
+                    outputRelativePath: join(nextBasename, "schema"),
```
fabioz, line 211: Question (it seems fine for me, just double-checking): if there are multiple input files, only a single `schema` is actually expected?

mstn: yes, the schema is the commonality across all bots. It works just like aav11.

PR #888, opened 2023-02-15
CONV-209 add metrics for converter

mstn:
It collects anonymous stats on converter usage.


mstn:
FYI @kariharju added new metrics. Is it ok?

fabioz:
Seems good to me. Just merged.

PR #890, opened 2023-02-16
CONV-210 show converter version in UI

mstn:


fabioz (changes requested):
The idea from the data is that it should store data from a previous invocation to restore afterwards, so, I don't think using it to put a `latestVersion` is the best way to go about it.

It'd be probably better to have a custom message sent from VSCode to the html to set the version.

i.e.: in the `converter.html` search for `window.addEventListener("message", (event) => {`, that's the place that processes messages received from VSCode.

Add a handler for a `setConverterVersion` which will update the UI accordingly.

Then, you can call: 

`panel.webview.postMessage({ command: "setConverterVersion", "version": converterVersion});`

when you have the version.


Also, there's an issue in the code getting the version: if you have one version when it starts up and it asks for an upgrade, this version will not be upgraded accordingly.

This happens because in:

```
    const convertBundlePromise: Promise<{
        pathToExecutable: string;
        pathToConvertYaml?: string;
    }> = ensureConvertBundle();
    const converterVersion = await getConverterVersion();
```

The first promise is not awaited prior to getting the version, so, it gets the version prior to actually doing the update.

My suggestion is doing a `convertBundlePromise.then()` to get the version and set it in the panel (if the panel wasn't disposed in the meanwhile).

fabioz:
Closing for now.

Issue #910, opened 2023-04-06
Configure Robotidy with ignore_git_dir option


bhirsz:
One of our users has asked us to implement ``--ignore_git_dir`` option in the Robotidy. It does change the way we are looking for the configuration file - basically Robotidy searches for ``pyproject.toml`` file in common parent directory of all sources, and it's going up to parent directories until it reaches beginning of path or finds ``.git`` directory.

For example if you have C:/project/repo_x/nested/file.robot and tell Robotidy to format ``file.robot`` it will go up until it reaches ``repo_x`` (since it contains ``.git`` directory like any other git repo).

Some users have different setup - for example multiple repo in one, sub-repo etc and prefer Robotidy to keep searching for config file even if ``.git`` dir is found. That's why I will be adding ``--ignore_git_dir`` option in the 4.1 release. But since this option need to be supplied to Robotidy before reading the configuration file (so it can find it), it needs to be passed from cli or api. 

That's why I need a way to pass parameter from lsp configuration (from your file?) to Robotidy, ie we can do it there: https://github.com/robocorp/robotframework-lsp/blob/973ab1b2fc40ebaa7dcc90018f90eaa249b5c29c/robocorp-python-ls-core/src/robocorp_ls_core/robotidy_wrapper.py (transform_model accept kwargs that can be used to set configuration).

I am willing to create the PR, I just need to figure out how you're parsing the configuration in meantime. There will be no problem will backward compatiblity (we're passing this option as kwarg, so older Robotidy that doesn't support this option will simply ignore it).

fabioz:
This is now in master (and in a few minutes in a `1.10.2` pre-release).

PR #931, opened 2023-05-12
Passignore_git_dir flag to Robotidy

bhirsz:
Add ignore_git_dir option to keep searching for configuration file even if there is .git directory.

Relates #910

It will work on all Robotidy versions (thanks to ** kwargs in our implementation). It's not ideal implementation - it's better to supply flag value from the LSP configuration (default False) but I don't know where I can configuration options to LSP. It shouldn't affect users even with flag always set to True - if they don't have configuration file stored somewhere in parent directories.


fabioz:
Seems good to me. Thanks for the PR!

PR #935, opened 2023-06-01
A360 read adapters from folder

mstn:
Let's revamp the API idea, with a different name.

Adapters are RF keywords that "adapt" Gen 1 RPA activities to RF keywords. Here, we implement the basic structure for A360.

Since we want to keep UI as simple as possible, the idea is that adapter code is not handled by users. Instead, adapters are stored in `$ROBOCORP_HOME/rca/VENDOR/adapters` folder. The folder files are "merged" and keywords are used as adapters if they occur in a to-be-converted bot.

I removed the old code of API (aka `apiFolder`) just to keep the code cleaner.


fabioz:
This is now merged.

PR #948, opened 2023-06-29
Playwright Integration - Initial version

custompointofview:
Adding Playwright recorder command - install playwright drivers, running recorder


fabioz (changes requested):
I'm taking a look at it.

The first thing I noted is that you put playwright into the extension environment. The idea was not using the extension environment for playwright.

Rather it should work on the environment of the target robot (so, it'd only work if the user has playwright in the robot dependencies -- if the user doesn't it'd let him know that it won't work).
Also, you should do the implementation on python (that's where most of the extension logic is -- javascript code should have as little as possible for Robocorp Code).

This means you'll have 2 actions (one is the one bound in the javascript side and the other is the extension .internal version which does the heavy lifting on the python side).

On the Python side the action should receive the target robot and then do something as:

```
            for ep in self._pm.get_implementations(EPResolveInterpreter):
                interpreter_info: IInterpreterInfo = (
                    ep.get_interpreter_info_for_doc_uri(uris.from_fs_path(target_robot))
                )
                if interpreter_info is not None:
                   # Use this interpreter to create the subprocess that'll use playwright.
```

For the first version you can just rely on robocorp.browser and use directly `robocorp.browser.goto()` to create the page for you (on a second step we'll probably copy a bunch of code from there to bootstrap playwright with the same logic).

robocorp-code/bin/create_env/conda.yaml
```diff
@@ -8,4 +8,5 @@ dependencies:
     - pip=22.1.2 # https://pip.pypa.io/en/stable/news/
     - pip:
           - robocorp-inspector==0.9.1 # https://github.com/robocorp/inspector/blob/master/CHANGELOG.md
+          - playwright==1.35.0 # https://pypi.org/project/playwright/#history
```
fabioz, line 11: These changes in `conda.yaml` adding playwright should no longer be needed.

robocorp-code/src/robocorp_code/playwright.py
```diff
@@ -0,0 +1,98 @@
...
+                )
+                if interpreter_info is not None:
+                    from subprocess import check_output
+
+                    python_path = interpreter_info.get_python_exe()
+                    log.debug("Found python interpreter:", python_path)
+
+                    # make sure we have the drivers installed
+                    log.debug("Installing playwright drivers...")
+                    cmd = [python_path, "-m", "playwright", "install"]
```
fabioz, line 55: Going forward we should go with the same code as the browser lib where it first tries to launch and if not then it'll install it and go forward.

I think that you should create a separate `.py` file which does all that (i.e.: tries to make the launch and installs if it fails and uses the robocorp-related paths to install).

Then you just need to run that separate file.

robocorp-code/src/robocorp_code/playwright.py
```diff
@@ -0,0 +1,98 @@
...
+                    # open the playwright recorder
+                    log.debug("Opening playwright recorder...")
+                    cmd = [
+                        python_path,
+                        "-m",
+                        "playwright",
+                        "codegen",
+                        "demo.playwright.dev/todomvc",
+                    ]
+                    b_output = check_output(cmd, timeout=35)
```
fabioz, line 68: We should probably just launch and forget here (i.e.: the method should return even while the user is still interacting with the process). The stdout/stderr should be piped and threads should log what's written from that process to `stderr` (which is automatically tracked by the extension to add to the `OUTPUT`).

Since the idea is having a separate `.py` file to handle actually launching playwright, that file can write some custom message to know whether it was started properly (so that `success` can be sent in that case).

fabioz (commented):
Great work. All the scaffolding is in place now.

Now we just need to work a bit on the actual core logic for starting playwright ;)

PR #951, opened 2023-07-10
CONV - enable python as a target language

mstn:
We add a dropdown in order to allow the selection of Python as a target language.

1. For now RF is set as default. This will change soon, but for now we want to release this feature silently.
2. We don't show the dropdown for BP because the Python backend is In progress

### DIFF
--- a/robocorp-code/vscode-client/templates/converter.html
+++ b/robocorp-code/vscode-client/templates/converter.html
@@ -559,7 +559,7 @@
                 <br />
                 <select id="targetLanguage" onchange="onTargetLanguageChanged()">
                     <option value="PYTHON">Python</option>
-                    <option value="RF">Robotframework</option>
+                    <option value="RF">Robot Framework</option>
                 </select>
             </div>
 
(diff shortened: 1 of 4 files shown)


robocorp-code/vscode-client/templates/converter.html
```diff
@@ -512,6 +554,15 @@
                 <div id="inputFilesOrFolders"></div>
             </div>
 
+            <div id="targetLanguageDiv" class="divEntry">
+                <label>Target language:</label>
+                <br />
+                <select id="targetLanguage" onchange="onTargetLanguageChanged()">
+                    <option value="PYTHON">Python</option>
+                    <option value="RF">Robotframework</option>
```
fabioz, line 562: Minor: It should be `Robot Framework` (instead of `Robotframework`).

fabioz (changes requested):
The code isn't handling loading existing values properly so that it sets the target language when loading the existing data (i.e.: search for `oldState` in `conversionView.ts`).

Because of that the default target language is actually `undefined`.

PR #960, opened 2023-07-21
Fix for opening playwright rec on Win 11

custompointofview:


robocorp-code/src/robocorp_code/playwright/__playwright__main.py
```diff
@@ -114,6 +118,7 @@ def in_thread():
             # -- note: this may be particular to my machine (fabioz), but it
             # may also be related to VSCode + Windows 11 + Windows Defender + python
             _stdin_write(process, b"\n")
+            _stdin_write(process, b"\n")
```
fabioz, line 121: Did this actually make a difference? The first call to `_stdin_write` should close `process.stdin`, so, in theory this will never do anything...

robocorp-code/src/robocorp_code/playwright/__playwright__main.py
```diff
@@ -168,14 +173,15 @@ def launch_playwright() -> None:
...
+    print(f"Playwright recorder finished. returncode: {returncode}", flush=True)
 
 
 def launch() -> None:
     try:
         launch_playwright()
     except Exception as e:
-        print(f"Failed opening recorder: {e}", flush=True)
+        print(f"Failed opening recorder:", flush=True)
+        print(e.__repr__().encode("utf-8"), flush=True)
```
fabioz, line 184: A better print would be:

```
sys.stdout.buffer.write(f'{e}\n'.encode(sys.stdout.encoding, 'replace'))
```

robocorp-code/src/robocorp_code/playwright/__playwright__main.py
```diff
@@ -115,18 +125,10 @@ def in_thread():
...
-                if not contents:
-                    break
-                full_output.append(contents)
-                if "<launched>" in contents:
-                    launched_count += 1
-                    if launched_count == 2:
-                        # It launches the browser and the 2nd window.
-                        launched_event.set()
+            # launch event timer -> it will release the lock on the launch event
+            launched_event_timer.start()
```
fabioz, line 129: I have a question here: didn't `<launched>` appear in the output for you (which is why it needed the change)?

fabioz:
Note: I did rebase this PR already (did a small change raising the timeout to 200).

PR #973, opened 2023-08-29
Bump Inspector version to 0.10.2 - rpaframework-core update

custompointofview:


fabioz:
There's some bad timing here because I'm about to do a change to the conda.yaml files in a branch too. Given that, I'm going to close this PR and make that change in the branch I'm working on.

PR #978, opened 2023-09-13
Inspector integration in VSCode

custompointofview:


robocorp-code/vscode-client/src/inspector/inspectorView.ts
```diff
@@ -0,0 +1,77 @@
...
+import * as vscode from "vscode";
+
+import { getExtensionRelativeFile, verifyFileExists } from "../files";
+import { logError } from "../channel";
+import { getSelectedRobot } from "../viewsCommon";
+import { LocatorsMap } from "./types";
+import { IMessage, IMessageType } from "./protocols";
+
+export async function showInspectorUI(context: vscode.ExtensionContext) {
+    const panel = vscode.window.createWebviewPanel(
```
fabioz, line 17: As a note, with this API it can only be created in the center (customizable would be `Active` or `Beside` columns) -- with this approach `Beside` may be better?

To put it in some other pre-defined place (such as `panel` which is in the bottom) it needs to be added to the `package.json` (see `codegen/views.py`) and then a `vscode.WebviewViewProvider` needs to be used.

If that's what we want you can search for `robocorp.python.view.output` and use that as a base (that's the `ROBO TASKS OUTPUT`).

robocorp-code/vscode-client/src/inspector/inspectorView.ts
```diff
@@ -0,0 +1,77 @@
...
+    const robot = getSelectedRobot();
+    if (robot) {
+        let locatorJson = path.join(robot.robot.directory, "locators.json");
+        if (verifyFileExists(locatorJson, false)) {
+            vscode.workspace.openTextDocument(vscode.Uri.file(locatorJson)).then((document) => {
+                let text = document.getText();
+                panel.webview.html = getWebviewContent(JSON.parse(text) as LocatorsMap);
+            });
+        } else {
+            logError("locators.json.not.found", undefined, "");
```
fabioz, line 36: I guess that this should not be considered an error... If the `locators.json` is not there it shouldn't be a problem as it should be created right?

p.s.: The `logError` is something that'll end up in our errors sent from RCC, which is not what we want (the `errorCode` in this case should be something significant to find the source afterwards, not an empty string).

p.s.: If you just want to do a regular log, use `OUTPUT_CHANNEL.appendLine` -- this will make the message appear in `Views > OUTPUT > Robocorp Code`.

robocorp-code/vscode-client/src/inspector/inspectorView.ts
```diff
@@ -0,0 +1,77 @@
...
+        }
+    } else {
+        logError("robot.not.found", undefined, "");
```
fabioz, line 39: Same here (no logError -- it should be the same for all the other occurrences).

robocorp-code/vscode-client/templates/inspector-bk.html
```diff
@@ -0,0 +1,98 @@
+<!DOCTYPE html>
```
fabioz, line 1: I guess this file can be removed?

fabioz (changes requested):
Just a few comments. 

Also, it'd be nice if it can be rebased with the latest master...

I had to do some changes because I was having issues with `setup.py` and `playwright` so I ended up using poetry for `Robocorp Code` (I was actually considering just using rcc, but the lack of separation from dev to regular dependencies was an issue).

robocorp-code/vscode-client/src/inspector/inspectorView.ts
fabioz: p.s.: No need to do this now, just wanted to point that out.

robocorp-code/vscode-client/templates/inspector-bk.html
custompointofview: Yep. Done!

robocorp-code/vscode-client/src/inspector/inspectorView.ts
custompointofview: Took a note on this. 
I haven't yet decided where this should view should be located. I'll try some other things out and see how they look. 
For now `Beside` seems to be the most appropriate. :)

PR #1000, opened 2023-11-16
Windows Inspector UI

custompointofview:


robocorp-code/src/robocorp_code/inspector/windows/windows_inspector.py
```diff
@@ -286,10 +293,11 @@ def stop_highlight(self) -> None:
         """
         Stops highlighting matches.
         """
-        if self._state == _State.highlighting:
-            if self._element_inspector is not None:
-                self._element_inspector.stop_highlight()
-            self._state = _State.default
+        log.info("Win-WinInsp-StopHighlight:state", self._state)
```
fabioz, line 296: This isn't the proper fix, the proper fix is setting the state properly in start highlight.

i.e.: add the lines below:

```
        self.reset_to_default_state()
        self._state = _State.highlighting
```

to the `start_highlight` method.

PR #1012, opened 2023-12-11
Inspector Apps - Improvements & Bug Fixes

custompointofview:


robocorp-code/src/robocorp_code/inspector/windows/robocorp_windows/_inspect.py
```diff
@@ -771,12 +776,70 @@ def stop(self):
         self._stop_event.set()
 
 
+class _CursorListenerThread(threading.Thread):
```
fabioz, line 779: Humm, instead of adding a thread that'll track the cursor, have you considered adding a mouse listener to the tk canvas elements created when the mouse is over it? -- and when it is remove the highlight...

custompointofview: I did. I tried various implementations - hovering, mouse enter, etc. - this was the only implementation that had a good enough user experience. If this isn't liked by users, we can try the other variants.

PR #1016, opened 2023-12-19
Inspector Apps - Quick Fixes & Improvements

custompointofview:
- Web - code generation, value of the selector
- Windows - foreground moves cursor to center when foreground is called
- Single instance of the Inspector

## Example of generated code:
image


robocorp-code/src/robocorp_code/inspector/windows/robocorp_windows/__init__.py
```diff
@@ -120,6 +120,7 @@ def find_window(
     timeout: Optional[float] = None,
     wait_time: Optional[float] = None,
     foreground: bool = True,
+    skip_cursor_centering: bool = False,
```
fabioz, line 123: Having a negation form is usually a bit strange and usually it's nicer to have the argument in a positive form.

i.e.:
```
move_cursor_to_center: bool = True
```

custompointofview: OK. Let me make the change.

fabioz (approved):

kariharju (approved):

PR #1018, opened 2023-12-23
Improve Configuration UI

yne:
This PR:
- Group configuration by categories
- Allow simple array and object modification from UI using the VScode "Add Item" action

How:
- By declaring an array of categories (each category contain the same "properties" values as before)
- By adding an `additionalProperties` attribute on object/list configurations params
- By updating the relevant intellij/package python generator

## Preview:
Untitled (1)

## Verification

Because of the new indentation level, this PR has a lot of changed lines,
In order to check that all attributes are still here from the previous package.json we can
```bash
diff \
<(jq '.contributes.configuration[].properties | keys[]' < package.json | sort)\
<(jq '.contributes.configuration.properties|keys[]' < package_ori.json | sort)

```


fabioz:
Sorry for the really late reply (I just saw this PR now).

Actually the `package.json` is generated from running `robotframework-lsp/codegen/codegen_package.py`, so, it cannot be changed directly (you need to update the code that generates it...).

yne:
@fabioz : I've updated the relevant generator in 617dd1dba7235e0bff288054fd86fde80fa876d0 then run a pass of generation that I've put in a 3rd commit 7f9fadc3d3c34f6d7137f0e36169d8f7a16c3d0d 

Tell me if I forgot any other files :)

yne:
closing for lack of reply

PR #1019, opened 2023-12-25
Inspector Improvements & New Features

custompointofview:
Web Recorder

- Feature - Handling iFrames or give warning about their existence on page + Detect iFrames & elements inside iFrames
- Feature - Sharable locator sets - Copy Locators JSON
- Feature - “Installing” status before “Stop recording”
- Feature - Make sure all generated code works with Playwright (updates from the Playwright API)
- Fix - Default wait_for_selector not working anymore
- Feature - Suggest all possible actions for the Web Locators (some might be not valid enough)
- Fix - UI/UX changes after frame_locator is introduced
- Fix/Feature - Rename Recorders to Inspectors
- Fix - Remove Java teaser from Commands Palette & Side Bar
- Fix - Screenshots - white on white text & other failures
- Feature - Add the Test Locator (Validate Strategy) for Action / Locator
- Feature - Add URL input to start the Browser page from a specific URL
- Feature - Generate Action for [ PAGE . GO TO ] = detect URL change and generate action
- Feature - Switch to the selected Inspector App every time a command is issued

Manager
- Fix/Feature - Rethink Locator details & messages - especially in the Manager state

Windows Recorder
- Feature - Change to the window when application is selected:  https://linear.app/robocorp/issue/INSP-122/how-can-we-get-the-main-window-locator-for-the-user
- Fix - Several fixes and consolidating

Backend (Robo Library + WebView BE)
- Most of the new Features and Fixes had BE integrations


robocorp-code/src/robocorp_code/inspector/web/_web_inspector.py
```diff
@@ -386,14 +400,18 @@ def inject_picker(self, reason: str, page=None) -> bool:
                     log.debug("page is None in inject_picker!")
                     return False
 
-            log.debug(f"Picker was not injected (injecting now. Reason: {reason}).")
-            page.evaluate(_load_resource("inspector.js"))
+            # we need to wait for the load state before injecting otherwise the evaluation of the picker crashes
+            # not finding the body where to inject the picker in
+            page.wait_for_load_state()
+            time.sleep(0.2)
```
fabioz, line 406: This sleep is a bit troublesome (if we need a sleep it may mean that on different machines it may not be enough).

I don't want to block the PR due to that, but we should try to fix things so that it's not needed...

fabioz (approved):

PR #1032, opened 2024-02-05
Image Inspector Integration

custompointofview:
- integrating the new Image Inspector in VSCode
- removing the old implementation of the Inspector
- removing indications of RPA package


robocorp-code/src/robocorp_code/inspector/image/_image_inspector.py
```diff
@@ -0,0 +1,135 @@
...
+        self._stop_event = threading.Event()
+        self.bridge = bridge
+        self.on_pick = on_pick
+        self.confidence = confidence
+
+    def run(self) -> None:
+        try:
+            self._run()
+        except Exception:
+            import traceback
```
fabioz, line 32: Should use `log.exception()` here.

robocorp-code/bin/create_env/conda_vscode_darwin_amd64.yaml
```diff
@@ -10,8 +10,9 @@ dependencies:
     - msgspec=0.18.2
     - robocorp-truststore=0.8.0 # https://github.com/sethmlarson/truststore/blob/main/CHANGELOG.md
     - python.app=1.3 # https://anaconda.org/conda-forge/python.app/files
+    - pyscreeze==0.1.30 # VSCode Image Inspector dependency
+    - python-mss==9.0.1 # VSCode Image Inspector dependency
     - pip:
           - --use-feature=truststore
-          - robocorp-inspector==0.10.2 # https://github.com/robocorp/inspector/blob/master/CHANGELOG.md
```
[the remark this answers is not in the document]
fabioz: This may be the reason for the current failing tests.

robocorp-code/src/robocorp_code/inspector/image/_bridge.py
```diff
@@ -0,0 +1,187 @@
...
+    from .vendor.utils import base64_to_image
+
+    # making sure we remove the unnecessary composition
+    img = img.replace("data:image/png;base64,", "")
+
+    return base64_to_image(img)
+
+
+class ImageBridge:
+    from robocorp_ls_core.protocols import IEndPoint
```
fabioz, line 30: Minor: usually imports should not be inside a class declaration.

Reason: it's mostly the same as a top-level import but the import will now also be available in the class API (which I don't think is what you want in this case).

robocorp-code/src/robocorp_code/inspector/image/_bridge.py
```diff
@@ -0,0 +1,187 @@
...
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            stdin=subprocess.PIPE,
+            env=environ,
+            bufsize=0,
+        )
+        stdout, _ = self.snipping_process.communicate()
+
+        # If you want to capture the output, you can use stdout.decode('utf-8') to convert it to a string
+        output = stdout.decode("utf-8")
```
fabioz, line 70: Use `stdout.decode("utf-8", "replace")` here (so that it won't fail with a UnicodeDecodeError in case some non utf-8 char is printed).

robocorp-code/src/robocorp_code/inspector/image/_bridge.py
```diff
@@ -0,0 +1,187 @@
...
+            self.snipping_process.terminate()
+            self.snipping_process.kill()
+        self.logger.info("=== Process stopped")
+
+    def validate(
+        self,
+        image_base64: str,
+        confidence: Optional[int] = None,
+    ):
+        needle = _base64_to_image(image_base64)
```
fabioz, line 123: I believe the validation was the really slow part right? (because pyscreeze is pretty slow for that without opencv).

My suggestion would be not even implementing it at this point until we have the proper backend functionality (so, I'd just raise an exception here... you can leave the current implementation commented out as it does have the part related to grabbing monitor info and so forth).

robocorp-code/src/robocorp_code/inspector/image/_image_inspector.py
```diff
@@ -0,0 +1,135 @@
...
+        # internals
+        self._image_bridge = ImageBridge(endpoint=endpoint, logger=log)
+        # threads
+        self._picker_thread: Optional[_PickerThread] = None
+        self._validate_thread: Optional[_ValidateThread] = None
+
+    def start_pick(self, confidence: Optional[int] = None) -> None:
+        log.info(">>> Image - Start pick...")
+
+        self._picker_thread = _PickerThread(
```
fabioz, line 105: This class spawns threads but I don't think it handles the internal state properly at this point.

i.e.:

If there's an existing `self._picker_thread` I guess it should first stop that one and then start a new one and the `on_pick` should only be related to the current thread.

Also, I'm not sure if this is meant to be thread-safe or not. Given that this is handling a bunch of internal state, can you do the following:

If it must be thread-safe, you probably should use a `threading.RLock` for checking existing and stopping it / creating new version/starting it in this case (and any access to `self._picker_thread`).

If it doesn't need to be thread safe because you know you won't access things from multiple threads verify that only the same threads access the function calls here.

i.e.: something as `self._use_in_thread = threading.current_thread()` and then at the start of each method do something as `assert self._use_in_thread == threading.current_thread()`

robocorp-code/src/robocorp_code/inspector/image/_image_inspector.py
```diff
@@ -0,0 +1,135 @@
...
+    def stop_pick(self) -> None:
+        log.info(">>> Image - Stop pick...")
+        if self._picker_thread:
+            self._picker_thread._stop_event.set()
+        if self._validate_thread:
+            self._validate_thread._stop_event.set()
+        self._image_bridge.stop()
+
+    def validate(self, image_base64: str, confidence: Optional[int] = None) -> None:
+        log.info(">>> Image - Validate pick...")
```
fabioz, line 121: You can leave the validate scaffolding, but we shouldn't call it at this point...

robocorp-code/src/robocorp_code/inspector/inspector_api.py
```diff
@@ -486,6 +486,151 @@ def __call__(
         return {"success": True, "message": None, "result": None}
 
 
+####
+#### IMAGE COMMANDS
+####
+class _ImageInspectorThread(threading.Thread):
+    """
+    This is a little tricky:
```
fabioz, line 494: Update comment?

fabioz (commented):
Nice work!

Still added some comments. 

Functionality-wise, I'd say the only thing is that we should not validate anything until we actually have a proper backend where the confidence would be the same thing and it's actually fast (I didn't review the html, so, not sure if that's already the case, but the backend should probably just raise an exception if some code tries to do it at this point).

robocorp-code/src/robocorp_code/inspector/image/_image_inspector.py
```diff
@@ -0,0 +1,145 @@
...
+        self.on_pick = on_pick
+        self.confidence = confidence
+
+    def run(self) -> None:
+        try:
+            self._run()
+        except Exception as e:
+            import traceback
+
+            traceback.print_exc()
```
fabioz, line 34: the `import traceback`/`traceback.print_exc()` should be removed.

robocorp-code/src/robocorp_code/inspector/image/_image_inspector.py
```diff
@@ -0,0 +1,145 @@
...
+        self.confidence = confidence
+        self.on_validate = on_validate
+
+    def run(self) -> None:
+        try:
+            self._run()
+        except Exception:
+            import traceback
+
+            traceback.print_exc()
```
fabioz, line 66: the `import traceback`/`traceback.print_exc()` should be removed and `log.exception()` should be used here.

robocorp-code/src/robocorp_code/inspector/inspector_api.py
```diff
@@ -8,13 +7,14 @@
 from robocorp_ls_core.python_ls import PythonLanguageServer
 from robocorp_ls_core.robotframework_log import get_logger
 
+
 from robocorp_code.inspector.web._web_inspector import (
     PickedLocatorTypedDict,
     WebInspector,
 )
+from robocorp_code.inspector.windows.windows_inspector import WindowsInspector
```
fabioz, line 15: This change is the opposite of what I mentioned earlier.

i.e.: the idea is that the code should be at the top-level:

```
if typing.TYPE_CHECKING:
    from robocorp_code.inspector.windows.windows_inspector import WindowsInspector
    from robocorp_code.inspector.image._image_inspector import ImageInspector
```

and then only locally, where it's actually needed, local imports would be done (so, we don't even import it unless it's actually needed).

custompointofview: I've tried that, but the execution fails. 
It seems that that IF isn't valid at runtime and `typing.TYPE_CHECKING` is false so the imports don't execute and the code breaks. 
That is why I reverted this and did direct imports.

fabioz: This is the missing:

> and then only locally, where it's actually needed, local imports would be done (so, we don't even import it unless it's actually needed).

part.

i.e.: you do that at top-level and then in the functions which actually use the `ImageInspector` you have to add the import again (but then, just inside that function).

fabioz (approved):
LGTM! Nice job :)

Issue #1033, opened 2024-02-08
missing requirements files in robocorp-code folder for contribution


otemek:
**Describe the bug**
According to the instructions in [Contribution guide](https://github.com/robocorp/robotframework-lsp/blob/master/robotframework-ls/docs/contributing.md#codeenvironment) about preparing code/environment either on Windows or Linux, there are missing requirements files (test/dev) which I guess are required to install and know what packages we need :)
```
...
python -m pip install -r robocorp-code/tests/test_requirements.txt
python -m pip install -r robocorp-code/dev_requirements.txt
...
```
Can I ask for some clarification/update on what the requirements are?

fabioz:
Thanks for the heads up (the page really needs an update as `robocorp-code` changed to use `pyproject.toml` with `poetry`).

For the Robot Framework Language Server you should be able to change those 2 lines for:

```
cd robotframework-ls
python setup.py install
pip install -r tests/test_requirements.txt
cd ..
```

The other lines should still be valid. As a note, you can always refer to how things work in github actions:

https://github.com/robocorp/robotframework-lsp/blob/master/.github/workflows/tests-robotframework-lsp.yml

p.s.: PRs to fix the docs welcome ;)

otemek:
closing as (at least for now) some part of the contribution documentation was updated. thanks

PR #1034, opened 2024-02-09
#1033 contribution docs update

otemek:
- Updating requirements.txt location
- change site-packages location from static `3.8` to auto-discover/assignment python version - unless it's intentional to keep 3.8 then this change can be removed


fabioz:
Thank you for the improvement. Just merged.

PR #1037, opened 2024-02-11
always archive the test results

otemek:
How about archiving test results - always? Because right now if there are any failures - those are not uploaded and can be seen only in a browser/workflow step section


otemek:
close this PR as it require more verification  which will do on a for repo

PR #1040, opened 2024-02-13
Mmokko/java inspector backend

mmokko:
The java locator backend implementation.


robocorp-code/pyproject.toml
```diff
@@ -39,6 +39,8 @@ pillow = "^9.0"
 pywin32 = { version = ">=300,<304", platform = "win32", python = "!=3.8.1" }
 psutil = "^5.9.0"
 comtypes = "^1.1"
+java-access-bridge-wrapper = "^1.1.0"
```
fabioz, line 42: `java-access-bridge-wrapper` also needs to be added to the conda files in `/robocorp-code/bin/create_env/` (that's the actual production env, `pyproject.toml` is just for the dev env).

robocorp-code/src/robocorp_code/inspector/java/robocorp_java/_event_pump.py
```diff
@@ -0,0 +1,63 @@
...
+            # TODO: add backoff timer
+            if not self._pump_background():
+                time.sleep(0.01)
+
+    def stop(self):
+        self._quit_queue_loop.set()
+        self._jab_wrapper = None
+
+    def get_wrapper(self) -> JavaAccessBridgeWrapper:
+        return self._queue.get()
```
fabioz, line 63: This is a bit odd in that it'd return from the queue once but then if it's no longer there it'll block forever if called again.

My suggestion is change the `self._queue = queue.Queue()` for `self._future = futures.Future()` and then the `get_wrapper()` can do `return self._future.result()` and the `run` can do `self._future.set_result(...)`.

There should be no reason to have an `self._jab_wrapper` anymore (if you want to remove the reference to it at `stop`, you have to dispose of the `_future` though.

robocorp-code/src/robocorp_code/inspector/java/robocorp_java/_inspector.py
```diff
@@ -0,0 +1,92 @@
...
+    },
+)
+
+
+class ElementInspector:
+    def __init__(self) -> None:
+        self._context: Optional[ContextNode] = None
+
+    def _start_event_pump(func, *args, **kwargs):
+        def wrapper(self: "ElementInspector", *args, **kwargs):
```
fabioz: The way the code is structured, the `JavaAccessBridgeWrapper` is created in a thread and an event loop is continually run in that thread. i.e.:

```
            if PeekMessage(message, 0, 0, 0, REMOVE_FROM_QUEUE):
                TranslateMessage(message)
                log.debug("Dispatching msg={}".format(repr(message)))
                DispatchMessage(message)
                return True
```

Now, although the `JavaAccessBridgeWrapper` was created in that thread, afterwards it's not really used in that thread, it's only used in this thread... 

Given that, my questions are the following: 
1. why create the `JavaAccessBridgeWrapper` in that thread in the first place? (i.e.: why not create it in this thread as that thread will just keep running that loop)
2. Isn't it dangerous having that loop running in a thread while having this thread manipulate the `JavaAccessBridgeWrapper`? (i.e.: usually UI's are single-threaded and accessing things from multiple threads could have bad repercussions)

mmokko: Jab wrapper needs to be part of the thread that pumps the window events, otherwise the API would stall and never return anything. Also we cannot call the API in that thread as we need to pump the events for the API to return results.

fabioz: Ok, (seems fair that the `JavaAccessBridgeWrapper` is in the same thread that's running the event loop). I guess everything is Ok then if the APIs used from this thread are thread safe :)

robocorp-code/bin/create_env/conda_vscode_windows_amd64.yaml
```diff
@@ -15,3 +15,4 @@ dependencies:
           - robocorp-inspector==0.10.2 # https://github.com/robocorp/inspector/blob/master/CHANGELOG.md
           - robotframework==5.0.1 # https://github.com/robotframework/robotframework/blob/master/doc/releasenotes/rf-5.0.1.rst
           - playwright==1.37.0
+          - java-access-bridge-wrapper=="1.1.0"
```
fabioz, line 18: Is this windows only?

PR #1043, opened 2024-02-21
Web Inspector fixes & improvements

custompointofview:
- add the `isolated` environment to the browser configuration every time the browser is initialised and when it's getting installed
- fixing the issues on MacOS with the browser getting stuck until user restarts VSCode
- new feature of changing the viewport resolution from the Web Inspector UI
- updated the vendored `robocorp-browser` to the latest version


robocorp-code/src/robocorp_code/playwright/robocorp_browser/__init__.py
```diff
@@ -208,20 +254,30 @@ def screenshot(
     Returns:
         The bytes from the screenshot.
     """
-    # import base64
+    import base64
+
+    from robocorp import log
```
fabioz, line 259: We should not import `robocorp.log` here (it's one of the reasons for this fork).

fabioz (commented):
Missing entry to robocorp-code/docs/CHANGELOG.md too...

Issue #1031, opened 2024-02-02
Load Variables From Variable File


BranislavHatala:
It would be useful if we could load variables for code-completion and code-analysis from variable files.
It would be nice if this supported variable files that uses function and possibly parameters.

Issue #1005, opened 2023-12-01
--variablefile support for RFLS?


Frivicktz:
**Is your feature request related to a problem? Please describe.**
My workplace are handling variables through a python file that is being ran before every run through the --variablefile command line option. This makes it impossible to verify anything that has to do with variables through RFLS. 

**Describe the solution you'd like**
--variablefile support for the language server. It is already supporting --argumentfile so it should be able to support --variablefile as well.

**Describe alternatives you've considered**
An alternative would be to add it through the imports in the file. But the --variablefile at the start is much more fluent.

SimonTate:
I would like this as well. I'm currently attempting to write test code that use a different --variablefile per DUT, and this is not possible to test with VSCode/ lsp right now.

Would be a great addition to simplify things to get this closer to just running robot command line.

Issue #953, opened 2023-07-11
possibility to use a python variable file for linting


linnfaa:
**Is your feature request related to a problem? Please describe.**
I would like to have the possibility to use a python variable file when linting with robot framework lsp in VSC. Currently there is no way to get variables from a python variable file to lsp.
https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#variable-files

**Describe the solution you'd like**
I think it would be nice to have an option for linting with a python variable file in the same way as from a argument file.

**Describe alternatives you've considered**
The alternatives I have found now are to lint variables by defining them in settings.json. This will make it problematic to change between different sets of variables (that has the same variables but with different values). Another way is to use argument file but i would rather use a python file.

Vikto1230:
I am missing this option aswell!

PR #1045, opened 2024-02-23
possibility to use a python variable file for linting

viktor1230:
Correction according to issues: #953 #1031 #1005


robocorp-python-ls-core/src/robocorp_ls_core/libs/robocop_lib/robocop/config.py
```diff
@@ -13,6 +13,8 @@
 from robocop.exceptions import (
     ArgumentFileNotFoundError,
     CircularArgumentFileError,
+    VariableFileNotFoundError,
```
fabioz, line 16: The files under `robocorp_ls_core/libs/robocop_lib` are vendored files from `robocop` (so, they shouldn't really be used nor imported inside of the language server aside from just from its main public API to collect errors).

See the readme under `robocorp_ls_core/libs/robocop_lib` if an update is needed on how to update it (but if it has to be updated, it should be in its own PR).

robotframework-ls/src/robotframework_ls/impl/variales_from_variable_file.py
```diff
@@ -0,0 +1,90 @@
+from typing import Optional, Tuple
+import os
+from robotframework_ls.impl.protocols import IVariableFound
+from robocorp_ls_core.robotframework_log import get_logger
+
+log = get_logger(__name__)
+
+
+class VariablesFromVariablesFileLoader:
```
fabioz, line 9: There should be at least one new test case for this one.

See: robotframework_ls_tests.test_variables_from_arguments_file.test_variables_from_arguments_file as a base on how it should look like.

robotframework-ls/package.json
```diff
@@ -441,6 +441,11 @@
                     "default": "",
                     "description": "Load variables for code-completion and code-analysis based on an arguments file. Multiple files accepted by separating with a comma."
                 },
+		"robot.loadVariablesFromVariablesFile": {
```
fabioz, line 444: There should be at least one test-case for this setting.

See robotframework_ls_tests.test_vscode_robot.test_variables_completions_integrated_using_config as a base.

fabioz (changes requested):
Thank you for the PR! Just a few notes added ;)

viktor1230:
Realized i messed up argument files, will try to correct according to this and your comments. Thanks.

robocorp-python-ls-core/src/robocorp_ls_core/libs/robocop_lib/robocop/config.py
viktor1230: Thank you for your comments. I have corrected the other ones but I don't understand this one as I am not familiar with vendored files. Should this not be included here?

fabioz: Well, the files in `robocorp_ls_core/libs/robocop_lib` are actually from the 'https://pypi.org/project/robotframework-robocop/` project, so, we should not do any manual change on those files. 

If they need to be updated you need to follow the instructions from: https://github.com/robocorp/robotframework-lsp/blob/master/robocorp-python-ls-core/src/robocorp_ls_core/libs/robocop_lib/README.md (but as I see no updates in the `.dist-info` with a version update, I believe this wasn't done...).

fabioz: As a note, any changes under `https://github.com/robocorp/robotframework-lsp/blob/master/robocorp-python-ls-core/src/robocorp_ls_core/libs/robocop_lib/` must be reverted in this PR...

fabioz: If we need something from them, we should ask the project maintainers to update it and then we can update the vendored files.

viktor1230: Thanks for your explanation. I reverted the changes.

fabioz:
Waiting for the tests to be fixed... there are typing issues with mypy and `"ImportError: cannot import name 'OPTION_ROBOT_VARIABLES_LOAD_FROM_VARIABLES_FILE' from 'robotframework_ls.impl.robot_lsp_constants'`...

viktor1230:
I'm sorry it took some time, but now I have attempted to fix according to tests.

viktor1230:
I believe I have fixed some of the failing tests. However there is still a few left that I don't really can get my head around, test_rcc_cloud in robocorp for example. How may my changes have affected that?

viktor1230:
@fabioz Do you mind triggering the tests again?

viktor1230:
Are the fails that are left related to my changes?

robotframework-ls/src/robotframework_ls/impl/variables_from_variable_file.py
```diff
@@ -0,0 +1,91 @@
...
+        multiline = ""
+        multiple_lines = 0
+        for lineno, line in enumerate(content.splitlines()):
+            line = line.strip()
+
+            if line.startswith("#"):
+                continue
+
+            if "#" in line:
+                line = line.split("#", 1)[0]
```
fabioz, line 32: Is this correct? Can't the `#` be escaped? Doesn't Robot Framework itself have some API which could be used to read this file?

i.e.: I think the current parsing can be pretty brittle (I'm ok in merging as is though if you don't want to pursue it further).

fabioz:
> Are the fails that are left related to my changes?

No, I don't think so (just waiting for answer to my previous comment to decide whether to merge as is or not).

robotframework-ls/src/robotframework_ls/impl/variables_from_variable_file.py
viktor1230: I have looked around without finding an API which can do the parsing. I noticed that comments in multiline variable, dict etc. screwed things up. As far as i have tested and used it this works for now.

fabioz: I think this is actually a regular python file.

You should be able to use: `robotframework_ls.impl.variable_completions_from_py.collect_variables_from_python_ast` to collect the variables from it.

viktor1230: How can I call that function when I have path as variable?

viktor1230: @fabioz Should i create a RobotDocument of the file to be able to parse it with collect_variables_from_python_ast?

viktor1230: I think I have corrected according to comments now. Could you please retrigger tests?

viktor1230: @fabioz I think it might pass the tests now, do you mind triggering it?

viktor1230:
@fabioz I'm sorry for bugging you. But i think the tests might pass now, if you could trigger them?

fabioz:
Thank you for the PR!

This is now merged in https://github.com/robocorp/robotframework-lsp/commit/afda578e5293fbc823e2894eef31e5ccad9acfa6

Note: I did fix the `test_variables_from_variables_file` which was broken and made an additional commit to fix the code generation (https://github.com/robocorp/robotframework-lsp/commit/51dbf3755cf7cda2f063cf33115a038e4c2bffb7).

I'll create a new release with this shortly.

PR #1066, opened 2024-05-01
Improvements to the rebuild env cycle

custompointofview:
Updated rcc. Fixed issue in the Caching system for the RCC Python Env. Added command for rebuilding the rcc environment. Added shortcut to the Activities for the task/action package. Added listeners for workspace files that trigger question for user to rebuild env or refresh view


robocorp-code/vscode-client/src/activities.ts
```diff
@@ -156,19 +156,28 @@ export async function resolveInterpreter(targetRobot: string): Promise<ActionRes
...
                 roboCommands.ROBOCORP_RESOLVE_INTERPRETER,
                 {
                     "target_robot": targetRobot,
                 }
             );
-            return result;
+            if (interpreter === null || (typeof interpreter === "string" && interpreter === "null")) {
+                throw Error("Interpreter not found");
+            }
+            vscode.window.showInformationMessage(`Environment built & cached. Python interpreter loaded.`);
```
fabioz, line 176: A message should not be shown if everything is correct here. You may want to move it to `ROBOCORP_PACKAGE_ENVIRONMENT_REBUILD` though.

robocorp-code/vscode-client/src/activities.ts
```diff
@@ -156,19 +156,28 @@ export async function resolveInterpreter(targetRobot: string): Promise<ActionRes
...
+            return interpreter;
         } catch (error) {
             logError("Error resolving interpreter.", error, "ACT_RESOLVE_INTERPRETER");
+            vscode.window.showErrorMessage(`Error resolving interpreter: ${error}`);
```
fabioz, line 180: At this point, as this can be called many times from different places, I think we shouldn't show an error here either (may be wanted in ROBOCORP_PACKAGE_ENVIRONMENT_REBUILD though).

robocorp-code/vscode-client/src/extension.ts
```diff
@@ -723,6 +726,19 @@ export async function doActivate(context: ExtensionContext, C: CommandRegistry)
...
+        const call = async () => {
+            const selected = await listAndAskRobotSelection(
+                "Please select the Task/Action Package for which you'd like to rebuild the environment",
+                "Unable to continue because no Action Package was found in the workspace.",
+                { showActionPackages: true, showTaskPackages: false }
+            );
+
+            await resolveInterpreter(selected.filePath);
+        };
+        call();
```
fabioz, line 739: Seems a bit strange to create an internal function here and call it exactly once (tip: you can make the outer one `async` by using adding `async` to it -- i.e.: `async () => `).

robocorp-code/vscode-client/src/pythonExtIntegration.ts
```diff
@@ -45,24 +62,54 @@ export async function autoUpdateInterpreter(docUri: Uri) {
...
-                if (docUri) {
-                    await autoUpdateInterpreter(docUri);
+export async function installWorkspaceWatcher(context: ExtensionContext) {
+    // listen for document changes
+    const checkIfFilesHaveChanged = workspace.onDidChangeTextDocument(async (event) => {
+        let docURI = event.document.uri;
+        if (event.document.isDirty && (isEnvironmentFile(docURI.fsPath) || isPythonFile(docURI.fsPath))) {
+            dirtyWorkspaceFiles.add(docURI.fsPath);
+        }
+    });
```
fabioz, line 72: The old auto-update on changes is still needed.

The use-case for it is having multiple tasks where each task has a different environment (so, by changing the open file the "current" interpreter must be changed).

fabioz (changes requested):

robocorp-code/vscode-client/src/pythonExtIntegration.ts
custompointofview: Got it. 
Will update.

robocorp-code/vscode-client/src/rccTerminal.ts
```diff
@@ -77,22 +77,58 @@ export async function createRccTerminal(robotInfo: LocalRobotMetadataInfo) {
...
+            // send setup commands to the terminal
+            if (process.platform.toString() === "win32") {
+                const terminal = window.createTerminal({
+                    name: robotInfo.name + " Package environment",
+                    env: env,
+                    cwd: pathModule.dirname(robotInfo.filePath),
+                    message: "Robocorp Code Package Activated Interpreter (Python Environment)",
+                    shellPath: "C:\\Windows\\System32\\cmd.exe",
+                });
+                const varsFilePath = pathModule.join(env.RCC_HOLOTREE_SPACE_ROOT, "environment_vars.bat");
```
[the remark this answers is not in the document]
custompointofview: Yes. It doesn't work on all environments. 
Even if the environment variables are injected properly into the terminal, there might be startup scripts that can execute when the shell is launched and the PATH (+ others) vars can be changed.
This way we ensure that the PY environment is activated properly.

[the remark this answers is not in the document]
fabioz: Can you add to the code a comment with the actual use case which triggered this and how to reproduce it?

This is pretty surprising, so, it warrants a proper explanation in the code...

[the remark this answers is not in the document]
custompointofview: Done.

robocorp-code/src/robocorp_code/rcc.py
```diff
@@ -1046,6 +1046,7 @@ def return_failure(msg: Optional[str]) -> ActionResult[IRobotYamlEnvInfo]:
             str(conda_yaml_path),
         ]
         args.append("--json")
+        args.append("--no-retry-build")
```
fabioz, line 1049: Can you also add `--no-pyc-management` here? It's something that should've been done quite a while ago too...

robocorp-code/vscode-client/src/pythonExtIntegration.ts
```diff
@@ -43,26 +60,77 @@ export async function autoUpdateInterpreter(docUri: Uri) {
...
+                dirtyWorkspaceFiles.delete(docURI.fsPath);
+                // if environment file has changed, let's ask the user if he wants to update the env
+                if (isEnvironmentFile(docURI.fsPath)) {
+                    window
+                        .showInformationMessage(
+                            "Changes were detected in the package configuration. Would you like to rebuild the environment?",
+                            "Yes",
+                            "No"
+                        )
+                        .then(async (selection) => {
```
fabioz, line 108: Note: could've used `selection = await window.showInformationMessage` instead of callbacks.

robocorp-code/vscode-client/src/pythonExtIntegration.ts
```diff
@@ -43,26 +60,77 @@ export async function autoUpdateInterpreter(docUri: Uri) {
...
+
+    // listen for document changes and mark targeted files as dirty
+    const checkIfFilesHaveChanged = workspace.onDidChangeTextDocument(async (event) => {
+        let docURI = event.document.uri;
+        if (event.document.isDirty && (isEnvironmentFile(docURI.fsPath) || isPythonFile(docURI.fsPath))) {
+            dirtyWorkspaceFiles.add(docURI.fsPath);
+        }
+    });
+
+    // listen for when documents are about to be saved
```
fabioz, line 88: This comment: `// listen for when documents are about to be saved` is no longer true here.

robocorp-code/vscode-client/src/pythonExtIntegration.ts
```diff
@@ -43,26 +60,77 @@ export async function autoUpdateInterpreter(docUri: Uri) {
...
+    const checkIfFilesWillBeSaved = workspace.onDidSaveTextDocument(async (document) => {
+        try {
+            let docURI = document.uri;
+            if (
+                docURI &&
+                dirtyWorkspaceFiles.has(docURI.fsPath) &&
```
fabioz, line 94: I'm not sure I follow the need for `dirtyWorkspaceFiles`... isn't just listening for the save and checking `isEnvironmentFile(docURI.fsPath) || isPythonFile(docURI.fsPath)` enough? 

I mean the save should only happen when an actual file was saved (in which case it must've changed in the meanwhile, no?) -- if that's the case the

custompointofview: One can still trigger the save listener even if the file hasn't been modified. That's why I've added the dirty files store so that we trigger only when the file has been modified.

PR #1087, opened 2024-12-12
IGNORE: Fix VAR syntax support in Robot Framework Language Server

BrunoMoraes-Z:
Add support for the new VAR syntax from Robot Framework 7.0 in Robot Framework Language Server.

* Update `robotframework-ls/src/robotframework_ls/impl/variable_completions.py` to handle the new VAR syntax in the `complete` and `_collect_resource_imports_variables` functions.
* Update `robotframework-ls/src/robotframework_ls/impl/variable_resolve.py` to support the new VAR syntax in the `token_value_and_unresolved_resolving_variables` and `_resolve_tok_var` functions.
* Update `robotframework-ls/src/robotframework_ls/impl/protocols.py` to include the new VAR syntax in the `IRobotToken` and `IVariableFound` interfaces.
* Update `robotframework-ls/src/robotframework_ls/impl/doc_highlight.py` to recognize and highlight the new VAR syntax in the `_highlight_variables` and `doc_highlight` functions.


BrunoMoraes-Z:
Sorry, I ended up making a miss click, I'm still testing whether the solution would work correctly.

fabioz:
No problems. As a note, the current failures related to the latest RF can be seen by running the tests with the master version of RF. Ideally you'd also create one test for it (you can use https://github.com/robocorp/robotframework-lsp/blob/master/robotframework-ls/tests/robotframework_ls_tests/completions/test_variables_completions.py as a base).

The language server was created with TDD (test driven development), so, you should be able to find tests for most of the things there and to support the latest version of RF is almost a matter of checking the errors vs that version of RF and fixing them.

PR #1095, opened 2025-06-08
Fix/dictionary completions test

BrunoMoraes-Z:


BrunoMoraes-Z:
I'm still working on some things, when I'm more stable then I intend to raise a PR.

